PR #2, opened 2020-07-02
Hierarchical command system support

artemsen:
- Old scripts are migrated into new format;
- Added command parameters support for sudoers files;
- Added bash autocomplete for commands (names only);

Signed-off-by: Artem Senichev <a.senichev@yadro.com>


README.md
```diff
@@ -10,22 +10,43 @@ to system resources that belong to root.
...
 
+## Top level (script)
 The script must contain description used for constructing system-wide help.
 Description is a comment inside the script started with identifier `CLI:`,
 for example:
 `# CLI: One-line comment that describes command purpose`
+
+## Function declaration
+
+### Executing privileges
```
AlexanderAmelkin: I believe it must be "Execution", not "Executing" here and in the end of the first paragraph. "Executing" looks like a gerund, while you obviously meant nouns as adjectives in both those places.

clicmd
```diff
@@ -1,25 +1,114 @@
...
+    exec sudo ${USER_CMD} ${sub_path//_/ } "$@"
+  fi
+
+  # execute subcommand
+  ${cmd} "$@"
+}
+
+# Print help for specified command.
+function print_help {
+  sed -n "/^#\\s\\+@doc\\s\\+${1}\\s*$/,/^function/ s/^#\\s*//p" "${REAL_CMD}" | tail +2
```
AlexanderAmelkin: Better use `tail -n+2`, it is more portable. Not all `tail`s can handle the `+2` without `-n`. My full-fledged GNU tail 8.25 can not, for instance. Busybox version can handle both variants, so it's better to use `-n+2`, I think.

commands/bmc
```diff
@@ -0,0 +1,252 @@
...
+                                        $((sec % 86400 / 3600)) \
+                                        $((sec % 3600 / 60)) \
+                                        $((sec % 60))
+  echo -n "Last BMC start was on "
+  format_date "${compact}" "${utc}" "${up}"
+}
+
+# @sudo cmd_info_version admin,operator,user
+# @doc cmd_info_version
+# Show version of BMC and BIOS/UEFI
```
AlexanderAmelkin, line 59: Do you propose we print "BIOS/UEFI" and have the `--bios` option even for OpenPOWER?

commands/bmc
```diff
@@ -0,0 +1,252 @@
...
+function cmd_datetime {
+  subcommand "$@"
+}
+
+# @sudo cmd_datetime_set admin
+# @doc cmd_datetime_set
+# Set system time manually
+#   YYYY-MM-DD-HH-mm-SS   Time to set
+function cmd_datetime_set {
+  if [[ $# -eq 0 ]] || [[ $1 =~ ^\d{4}-\d\d-\d\d-\d\d-\d\d-\d\d$ ]]; then
```
nest1ing, line 98: Hm... Looks strange. Try to run it several times with the different values ;)
Maybe it would better to use here an ISO format ("YYYY-MM-DD HH:mm:SS")?
Or it might be just 14 digits with any combinations of [` `, `-`, `:`] as delimiters or without ones.
It seems to me, there should be either well-known format or any possible combinations here.

commands/bmc
```diff
@@ -0,0 +1,252 @@
...
+    shift
+  done
+  format_date "${compact}" "${utc}" "$(date +%s)"
+}
+
+# @sudo cmd_datetime_ntpconfig admin
+# @doc cmd_datetime_ntpconfig
+# Configure NTP server connection
+function cmd_datetime_ntpconfig {
+  netconfig "$@"
```
nest1ing: Looks like some `netconfig` arguments are missing here.

commands/bmc
```diff
@@ -0,0 +1,252 @@
...
+}
+
+# @sudo cmd_ifconfig admin
+# @doc cmd_ifconfig
+# BMC network configuration
+function cmd_ifconfig {
+  netconfig "$@"
```
nest1ing, line 143: And here too

commands/health
```diff
@@ -0,0 +1,60 @@
...
+# Get list of system logs
+#function cmd_logs_list {
+#  not implemented
+#}
+
+# @sudo cmd_logs_show admin,operator
+# @doc cmd_logs_show
+# Get system logs
+function cmd_logs_show {
+  expect_noarg "$@"
```
nest1ing: `-f|--follow` is very useful.

commands/host.vesnin
```diff
@@ -0,0 +1 @@
+host.nicole
```
nest1ing, line 1: ???
If it is a dummy file, just leave it empty. Or just remove it from git.

commands/bmc
artemsen: I can't, this format is required by our internal doc (see VEGMAN CLI Proposal).

commands/bmc
artemsen: There are many arguments for netconfig and they all passed directly to the executable module.
Script handles such situations and does it in special way when internal help is not available.

commands/bmc
artemsen: When function doesn't contain help - all parameters are passed to netconfig.

commands/health
artemsen: Done according to the internal document.

commands/host.vesnin
artemsen: It is not a dummy file. It will be installed on vesnin machines (if we wish). We need to split host-specific handlers.

commands/bmc
AlexanderAmelkin: I think we can safely expand the proposal and add some extra formats.
I agree that delimiters must be optional and must be any of `[- /\\\.]` for date and any of `[- :\.]` for time.
I would also support omitting the seconds.

commands/bmc
```diff
@@ -0,0 +1,252 @@
...
+  subcommand "$@"
+}
+
+# @sudo cmd_datetime_set admin
+# @doc cmd_datetime_set
+# Set system time manually
+#   YYYY-MM-DD-HH-mm-SS   Time to set
+function cmd_datetime_set {
+  if [[ $# -eq 0 ]] || [[ $1 =~ ^\d{4}-\d\d-\d\d-\d\d-\d\d-\d\d$ ]]; then
+    echo "Please enter date and time in YYYY-MM-DD-HH-mm-SS format" >&2
```
AlexanderAmelkin, line 99: I also think that the proposal document missed the `UTC+X:XX` part in this hint text.
I think you've missed it in the code too.
I personally think that the word `UTC` is redundant and the arguments must be like this:
`datetime set YYYY-MM-DD-HH-mm-SS ±XX:YY`, the timezone is always UTC.
The hint text must be like this:
`Please enter date and time in YYYY-MM-DD-HH-mm-SS [±HH:mm] format, the optional offset is relative to UTC`

commands/bmc
AlexanderAmelkin: Yes, but this is `datetime ntpconfig` command and you're passing all of its arguments directly to `netconfig`. Are you sure that's what you meant to do?

commands/bmc
AlexanderAmelkin: Are we masking the system's standard `ifconfig` somehow?
I'm quite sure we also need to prohibit users from fiddling with `ip` directly.

commands/bmc
```diff
@@ -0,0 +1,252 @@
...
+
+  if [[ -n "${force}" ]]; then
+    echo -n "BMC will be rebooted gracefully,"
+  else
+    echo -n "BMC will be rebooted forcefully,"
+  fi
+  echo " all user sessions including this one will be"
+  echo "closed. Host will not be affected."
+
+  confirm
```
AlexanderAmelkin: To aid scripting, let's add the `-y` option to all commands asking for confirmation.

commands/health
AlexanderAmelkin: First of all, that's not exactly true. The document describes the `logs show` command as it is implemented in klish now. That is, "Show selected log". `journalctl` is just one of the dreport logs. Please implement `logs list` and `logs show <log>` as it is done (behavior-wise) in VESNIN 2.x

commands/health
```diff
@@ -0,0 +1,60 @@
...
+  journalctl --no-hostname --no-pager
+}
+
+# @sudo cmd_logs_export admin,operator
+# @doc cmd_logs_export
+# Export system logs
+function cmd_logs_export {
+  expect_noarg "$@"
+  local file="/tmp/bmc_$(date +'%Y%m%d_%H%M%S').log"
+  journalctl --no-hostname --no-pager --verbose > "${file}"
```
AlexanderAmelkin: Again, this is about exporting dreport logs (or a single one of them).

commands/host.nicole
```diff
@@ -0,0 +1,245 @@
...
+#
+# CLI: Host operations
+#
+
+# @sudo cmd_info admin,operator
+# @doc cmd_info
+# View the general host information
+function cmd_info {
+  expect_noarg "$@"
+  obmcutil status | sed 's/xyz.*\.//'
```
AlexanderAmelkin, line 12: I believe we did more parsing than this in klish. `obmcutil` gives quite a non-friendly output.
Please make it more user-friendly as we did in VESNIN. Please also check other exising commands for any post-processing of the raw output they may be doing there. Let's not discard what have already been done.

commands/host.nicole
```diff
@@ -0,0 +1,245 @@
...
+
+# @doc cmd_boot_next
+# Choose boot source for the next boot
+#   -l, --local
+#   -p, --pxe
+#   -s, --san
+#   -u, --usb
+#   -m, --media
+#function cmd_boot_next {
+#  not implemented
```
AlexanderAmelkin, line 29: This is quite easily implementable via dbus, see:

1. https://github.com/openbmc/phosphor-dbus-interfaces/blob/master/xyz/openbmc_project/Control/Boot/ (`Mode` and `Source` interfaces)
2. https://github.com/openbmc/phosphor-dbus-interfaces/blob/master/xyz/openbmc_project/Object/ (`Enable` interface)
3. /xyz/openbmc_project/control/host0/boot/one_time
4. https://github.com/openbmc/phosphor-host-ipmid/blob/master/chassishandler.cpp (`ipmi_chassis_set_sys_boot_options()`)

The easiest way is actually to include `ipmitool` package with `dbus` interface enabled, and then just use it. Intel does it that way for `s2600wf`.

commands/host.vesnin
AlexanderAmelkin: As far as I can see, you do machine switching inside the `host.nicole` script.
I would understand if you had a `host` script that would source a `host.$MACHINE` script with machine-specific commands and then would just call machine specific callbacks like `nmi_machine`, but doing direct machine switching in `host.nicole` and then having this dummy doesn't look neat.

I'd even expect `clicmd` to pre-load the machine-specific command file before running the main handler that could then check for availability of a machine-specific function (using `type`) and behave accordingly (e.g. report that the function is not supported for the machine).

install.sh
```diff
@@ -4,12 +4,24 @@
 # Installer.
 #
 
+THIS_DIR="$(realpath "$(dirname "$0")")"
 INSTALL_ROOT=""
 MACHINE="nicole"
```
AlexanderAmelkin, line 9: ??? How is this overridden for vesnin or vegman?

AlexanderAmelkin (changes requested):
I'm not yet done with the review, but I'm submitting it now for you to have something to work on. :)

commands/bmc
artemsen: I will do as soon as it will be described in the proposal.

commands/bmc
artemsen: Yes, these utils are in the /sbin directory and an end user doesn't have it in the PATH.

commands/bmc
artemsen: Yes, see the proposal.

install.sh
artemsen: Why do we need to override the source directory?

commands/bmc
AlexanderAmelkin: The proposal document is written for VEGMAN, which is x86. For OpenPOWER systems this will look strange. The document has already been updated for this as far as I know.

commands/bmc
AlexanderAmelkin: The proposal also requires support for UTC offset.

commands/health
artemsen: Do we really need to implement the third way to show uptime? :)
Vesnin calls dreport plugins which behavior depends on dreport library and we need to implement this library too. The second problem - dreport scripts doesn't intend to work under sudo and we have to audit every change in this project to be sure that there are no any security vulnerabilities.

Here the list of plugins:
E10diskusage       E1pnorversion      E30ipaddr          E5chassisstate
E10failedservices  E20audit           E30iplink          E5hostnamectl
E10journalpretty   E20elogall         E40fwprintenv      E5hoststate
E10meminfo         E20hostlogger      E50cpuinfo
E10top             E20inventory       E50uptime
E1osrelease        E25obmcconsole     E5bmcstate

Most of them are already implemented by other commands. And I don't understand how CPU info or Top relates to the `log` command.

AlexanderAmelkin: As you like to say, let's first change the document :)

AlexanderAmelkin (approved):
Anyway, I decided to accept this as it is and then fix what is wrong or doesn't correspond to the document on a per-issue basis.
Please rebase/resolve conflicts and let's merge this.

PR #10, opened 2020-10-05
user: Manage accounts via D-Bus calls

artemsen:
All operations related to account management are now performed via
D-Bus calls to xyz.openbmc_project.User.Manager.

Signed-off-by: Artem Senichev <a.senichev@yadro.com>

### DIFF
--- a/commands/user
+++ b/commands/user
@@ -12,11 +12,17 @@ GROUP_ADMIN="%GROUP_ADMIN%"
 GROUP_OPERATOR="%GROUP_OPERATOR%"
 GROUP_USER="%GROUP_USER%"
 GROUP_IPMI="ipmi"
-GROUP_SUDO="sudo"
 GROUP_REDFISH="redfish"
 GROUP_WEB="web"
+GROUP_SSH="ssh"
 
-# Lookup for group by role name
+# D-Bus entities to manage user accounts
+DBUS_SERVICE="xyz.openbmc_project.User.Manager"
+DBUS_ROOT_PATH="/xyz/openbmc_project/user"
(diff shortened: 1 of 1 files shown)


commands/user
```diff
@@ -12,11 +12,17 @@ GROUP_ADMIN="%GROUP_ADMIN%"
...
 GROUP_USER="%GROUP_USER%"
 GROUP_IPMI="ipmi"
-GROUP_SUDO="sudo"
 GROUP_REDFISH="redfish"
 GROUP_WEB="web"
+GROUP_SSH="ssh"
 
-# Lookup for group by role name
+# D-Bus entities to manage user accounts
+DBUS_SERVICE="xyz.openbmc_project.User.Manager"
```
nest1ing, line 20: The service name can be changed in the future.
But the interface name will be kept.
So, I think it would be more correct write:
```
DBUS_SERVICE="$(mapper get-service ${DBUS_ROOT_PATH})"
DBUS_IFACE="xyz.openbmc_project.User.Manager"
```
And then you should fix the variables usage in the code.

artemsen: It doesn't work as expected, actually. I get wrong service:
```
qemu:~ # mapper get-service /xyz/openbmc_project/user
xyz.openbmc_project.LDAP.PrivilegeMapper
```
Or did I miss something?

nest1ing: Wow, it is wrong. So, the more correct solution is the calling `GetObject` method  over D-Bus from the `ObjectMapper`.
But it might be hard to parse the answer in bash. So, I don't insist on it.

nest1ing (approved):

PR #12, opened 2020-11-10
bmc/datetime: add time synchronization method

nest1ing:
This brings the interface to show/change the time synchronization
method.

### DIFF
--- a/commands/bmc.vegman
+++ b/commands/bmc.vegman
@@ -4,6 +4,11 @@
 # CLI: General BMC configuration
 #
 
+SETTINGS_SERVICE="xyz.openbmc_project.Settings"
+TIMESYNC_METHOD_PATH="/xyz/openbmc_project/time/sync_method"
+TIMESYNC_METHOD_IFACE="xyz.openbmc_project.Time.Synchronization"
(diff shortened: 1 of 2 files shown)


artemsen (approved):

commands/bmc.vegman
```diff
@@ -90,6 +95,48 @@ function cmd_datetime {
   subcommand "$@"
 }
 
+# @sudo cmd_datetime_method admin
+# @doc cmd_datetime_method
+# Change time synchronization method
+#  {manual|ntp}
+function cmd_datetime_method {
+  local method="$(busctl get-property ${SETTINGS_SERVICE} \
```
AlexanderAmelkin, line 103: We need to think about joining the common code and not copy-pasting it around many files.

artemsen: It is not so easy, actually.
I have already tried, but it either leads to a too complex solution or to very slow handling of commands.

AlexanderAmelkin: You do install anyway. Maybe have a '.common' file and just add it to the end of the 'installed' script for the target?

AlexanderAmelkin (approved):

PR #15, opened 2020-12-16
bmc/syslog: add implementation

nest1ing:
This brings an implementation of the remote logging settings for all
our platforms.

### DIFF
--- a/commands/bmc.nicole
+++ b/commands/bmc.nicole
@@ -9,6 +9,10 @@ TIMESYNC_METHOD_PATH="/xyz/openbmc_project/time/sync_method"
 TIMESYNC_METHOD_IFACE="xyz.openbmc_project.Time.Synchronization"
 TIMESYNC_METHOD_PROPERTY="TimeSyncMethod"
 
+SYSLOG_CONFIG_SERVICE="xyz.openbmc_project.Syslog.Config"
+SYSLOG_CONFIG_PATH="/xyz/openbmc_project/logging/config/remote"
+NETWORK_CLIENT_IFACE="xyz.openbmc_project.Network.Client"
+
 # Format date in a non-standard way
 function format_date {
   local compact="$1"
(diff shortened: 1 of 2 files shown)


commands/bmc.nicole
```diff
@@ -191,11 +195,62 @@ function cmd_ifconfig {
...
+        fi
+        echo
+    fi
+}
+
+# @sudo cmd_syslog_set admin
+# @doc cmd_syslog_set
+# Configure remote syslog server
+#  address[:port] - to change settings
+#  empty - to reset settings
```
artemsen, line 236: I would suggest to make a subcommand for resetting, lomething like `bmc syslog reset`.

nest1ing: done

commands/bmc.vegman
```diff
@@ -190,6 +194,71 @@ function cmd_ifconfig {
...
+        fi
+        echo
+    fi
+}
+
+# @sudo cmd_syslog_set admin
+# @doc cmd_syslog_set
+# Configure remote syslog server
+#  address[:port] - to change settings
+#  empty - to reset settings
```
artemsen, line 236: Not needed anymore.

AlexanderAmelkin: It still works as far as I can see

AlexanderAmelkin (approved):

commands/bmc.vegman
artemsen: I mean, lets have the one only official way to do it.

nest1ing: I think that `bmc syslog set` is shorter than `bmc syslog reset` ;)
And both ways work now.

artemsen: So, now you have to describe both ways in the Users' documentation ;-)

artemsen (approved):

alatarum (approved):

PR #16, opened 2021-01-25
vegman: tech: add FRU updating module

alatarum:
This brings implementation for service function to update particular FRU
fields. To do this the script restarts FruDevice daemon in special mode,
where number of parameters become writable.

Signed-off-by: Andrei Kartashev <a.kartashev@yadro.com>

### DIFF
--- a/commands/tech.vegman
+++ b/commands/tech.vegman
@@ -1,0 +1,161 @@
+#!/bin/bash -eu
+
+#
+# CLI: Service operations
+#
+
+# D-Bus object descriptions
+FRU_SYSTEMD_SERVICE="xyz.openbmc_project.FruDevice.service"
+FRU_DAEMON="fru-device"
+FRU_DBUS_SERVICE="xyz.openbmc_project.FruDevice"
+FRU_DBUS_OBJECT_TEMPLATE="\/xyz.*_Motherboard"
+FRU_DBUS_INTERFACE="xyz.openbmc_project.FruDevice"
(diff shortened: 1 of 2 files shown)


nest1ing:
Is it really need to be a part of the CLI project?
As I understood, this tool is very specific. It will be used by our manufacture guys only

nest1ing:
Oops, sorry. Missclick.

commands/tech.vegman
```diff
@@ -0,0 +1,161 @@
...
+    fi
+  done
+
+  echo "Parameters to be set in FRU:"
+  echo "    CHASSIS_PART_NUMBER:   ${fru_chassis_pnum}"
+  echo "    PRODUCT_PART_NUMBER:   ${fru_product_pnum}"
+  echo "    PRODUCT_PRODUCT_NAME:  ${fru_product_name}"
+  echo "    PRODUCT_SERIAL_NUMBER: ${fru_product_serial}"
+  if confirm; then
+    set_fru_field "CHASSIS_PART_NUMBER" "${fru_chassis_pnum}" || return 1
```
artemsen, line 149: `|| return 1` is not necessary, you have already set `-e` for entire script.

alatarum: IMHO, code should not rely on command line options.

AlexanderAmelkin: `set -e` is not a command line option. That's what is set in the script itself, and that's a reliable feature.

PR #17, opened 2021-02-02
vegman: Support virtual media and NBD client

ikmsk10:
Adds support bellow commands:

```bash
# host virtualmedia mount (--usb,--usb-ro,--hdd,--cdrom) </file/path>
# host virtualmedia umount [-y,--yes]
# bmc remoteimage connect (--nbd NUMBER) address[:port]
# bmc remoteimage disconnect [-y,--yes] --nbd NUMBER
```
    
This command set uses to allow access to VM functions for admin.

### DIFF
--- a/commands/bmc.vegman
+++ b/commands/bmc.vegman
@@ -353,3 +353,88 @@ function cmd_reboot {
   echo "Reboot the BMC..."
   reboot ${force}
 }
+
+# @doc cmd_remoteimage
+# Access to the remote side file
+function cmd_remoteimage {
+  subcommand "$@"
+}
+
+# @sudo cmd_remoteimage_connect admin
+# @doc cmd_remoteimage_connect
+# Connect to the remote server via the specified protocol
+#   --nbd <0..10>     use NBD protocol to connect the remote server with specified NBD device number
+#   address[:port]    remote server address
+function cmd_remoteimage_connect {
+    local nbd_device=""
+    local remote_point=""
+    local address=""
+    local port=""
+
+    while [[ $# -gt 0 ]]; do
+      case "${1}" in
+        --nbd)
+          if [[ $# -lt 2 ]]; then
+            echo "The NBD device number shall be specified" >&2
+            return 1
+          fi
+          nbd_device="/dev/nbd${2}"
+          shift
+          ;;
+        *)
+          if [[ -n "${remote_point}" ]]; then
+            abort_badarg "${1}"
+          else
+            remote_point="${1}"
+          fi
+          address="${remote_point%:*}"
+          port="${remote_point#*:}"
+          ;;
+      esac
+      shift
+    done
+
+    if [[ -z "${port}" ]]; then
+        port=10809
+    fi
+
+    if [[ -z "${address}" ]] || [[ -z "${port}" ]]; then
+      echo "The remote server address shall be specified." >&2
+      return 1
+    fi
+
+    echo "Connecting to NBD server ${remote_point}..."
+    nbd-client "${address}" "${port}" "${nbd_device}"
+}
+
+# @sudo cmd_remoteimage_disconnect admin
+# @doc cmd_remoteimage_disconnect
+# Dissconnect NBD session
+#   -y, --yes       do not ask for confirmation
+#   --nbd <0..10>   the number of NBD device
+function cmd_remoteimage_disconnect {
+  local yes=""
+  local nbd_device=""
+
+  while [[ $# -gt 0 ]]; do
+    case "${1}" in
+      -y | --yes) yes="y";;
+      --nbd)
+        if [[ $# -lt 2 ]] || [[ ${2:0:1} == "-" ]]; then
+          echo "The NBD device number shall be specified" >&2
+          return 1
+        fi
+        nbd_device="/dev/nbd${2}"
+        shift
+        ;;
+      *) abort_badarg "${1}";;
+    esac
+    shift
+  done
+
+  if [[ -z "${yes}" ]]; then
+    confirm
+  fi
+
+  nbd-client -d "${nbd_device}"
+}
--- a/commands/host.vegman
+++ b/commands/host.vegman
@@ -63,6 +63,71 @@ function cmd_power_status {
   ipmitool power status
 }
 
+# @doc cmd_virtualmedia
+# Access Virtual Media
+function cmd_virtualmedia {
+  subcommand "$@"
+}
+
+# @sudo cmd_virtualmedia_mount admin
+# @doc cmd_virtualmedia_mount
+# Mount a virtual media
+#  --usb    - the USB interface type
+#  --usb-ro - the USB interface type (read-only)
+#  --hdd    - the HDD interface type
+#  --cdrom  - the CD/DVD/BD-ROM interface type (read-only), based on file size
+#  FILE — path to image file
+function cmd_virtualmedia_mount {
+  local imagefile=""
+  local type=""
+
+  while [[ $# -gt 0 ]]; do
+    case "${1}" in
+      --usb) type="usb";;
+      --usb-ro) type="usb-ro";;
+      --hdd) type="hdd";;
+      --cdrom) type="cdrom";;
+      *)
+        if [[ -n "${imagefile}" ]]; then
+          abort_badarg "${1}"
+        else
+          imagefile="${1}"
+        fi
+        ;;
+    esac
+    shift
+  done
+
+  if [[ ! -e ${imagefile} ]] || [[ -d ${imagefile} ]]; then
+    echo "Please specify the path to image file" >&2
+    return 1
+  fi
+
+  usb-ctrl insert "usb.ms.cli" "${imagefile}" "${type}"
+}
+
+# @sudo cmd_virtualmedia_umount admin
+# @doc cmd_virtualmedia_umount
+# Unmount the virtual media
+#  -y, --yes  - does not ask for interactive confirmation
+function cmd_virtualmedia_umount {
+  local yes=""
+
+  while [[ $# -gt 0 ]]; do
+    case "${1}" in
+      -y | --yes) yes="y";;
+      *) abort_badarg "${1}";;
+    esac
+    shift
+  done
+
+  if [[ -z "${yes}" ]]; then
+    confirm
+  fi
+
+  usb-ctrl eject "usb.ms.cli"
+}
+
 # @doc cmd_console
 # Open virtual console
 function cmd_console {


commands/bmc.vegman
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash -eu
+#!/bin/bash -eu -x
```
artemsen: Forget to remove?

commands/bmc.vegman
```diff
@@ -353,3 +353,86 @@ function cmd_reboot {
...
+            remote_point="$1"
+          fi
+          address="${remote_point%:*}"
+          port="${remote_point#*:}"
+          ;;
+      esac
+      shift
+    done
+
+    if [[ "${port}" == "${remote_point}" ]] || [[ -z "${port}" ]]; then
```
artemsen: Too many quotes =)
I would check for empty port and then compare it:
```
    if [[ -z "${port}" ]] && [[ ${port} == ${remote_point} ]]; then
```

commands/bmc.vegman
```diff
@@ -353,3 +353,86 @@ function cmd_reboot {
...
+        port=10809
+    fi
+
+    if [[ -z "$address" ]] || [[ -z "$port" ]]; then
```
artemsen: Please, use single style for variables: either `${var}` or `$var`.

commands/host.vegman
```diff
@@ -63,6 +63,71 @@ function cmd_power_status {
...
+          abort_badarg "$1"
+        else
+          imagefile="$1"
+        fi
+        ;;
+    esac
+    shift
+  done
+
+  if [[ ! -f "$imagefile" ]] && [[ ! -e "$imagefile" ]]; then
```
artemsen: In bash, you need quotes only for `-z` or `-n`, other cases are covered by `[[`.

ikmsk10: Fixed

commands/bmc.vegman
ikmsk10: Fixed

commands/bmc.vegman
ikmsk10: Done

commands/bmc.vegman
ikmsk10: Yes, I forgot. Removed

artemsen (approved):

commands/bmc.vegman
```diff
@@ -353,3 +353,86 @@ function cmd_reboot {
...
   reboot ${force}
 }
+
+# @doc cmd_remoteimage
+# Access to the remote side file
+function cmd_remoteimage {
+  subcommand "$@"
+}
+
+# @sudo cmd_remoteimage_connect admin
```
nest1ing, line 363: The `operator` role allows their members to manage the host.
Shouldn't we also allow these commands for the operators?

ikmsk10: At the CLI Interface Proposal, the VirtualMedia command is allowed only for admin.
I have followed these recommendations, but of course, I think we might allow user access from the operator group.

commands/host.vegman
```diff
@@ -63,6 +63,71 @@ function cmd_power_status {
...
+          abort_badarg "${1}"
+        else
+          imagefile="${1}"
+        fi
+        ;;
+    esac
+    shift
+  done
+
+  if [[ ! -f ${imagefile} ]] && [[ ! -e ${imagefile} ]]; then
```
alatarum: why to test both -f and -e?
You also should test if type specified

commands/bmc.vegman
```diff
@@ -353,3 +353,86 @@ function cmd_reboot {
...
+# @doc cmd_remoteimage_connect
+# Connect to the remote server via the specified protocol
+#   --nbd <0..10>     use NBD protocol to connect the remote server with specified NBD device number
+#   address[:port]    remote server address
+function cmd_remoteimage_connect {
+    local nbd_device=
```
alatarum: local nbd_device=""

commands/bmc.vegman
```diff
@@ -353,3 +353,86 @@ function cmd_reboot {
...
+          else
+            remote_point="${1}"
+          fi
+          address="${remote_point%:*}"
+          port="${remote_point#*:}"
+          ;;
+      esac
+      shift
+    done
+
```
alatarum, line 396: I believe, you should test if remote_point empty here

commands/bmc.vegman
```diff
@@ -353,3 +353,86 @@ function cmd_reboot {
...
+    echo "Connecting to NBD server ${remote_point}..."
+    nbd-client "${address}" "${port}" "${nbd_device}"
+}
+
+# @sudo cmd_remoteimage_disconnect admin
+# @doc cmd_remoteimage_disconnect
+# Dissconnect NBD session
+#   -y, --yes   do not ask for confirmation
+#   NUMBER      the number of NBD device
```
alatarum: please, use same syntax as in cmd_remoteimage_connect (--nbd <0..10>)

commands/host.vegman
ikmsk10: Because the image file may be a UNIX device, which should be able to read.
The flag `-f` does not positively affect `/dev/nbdX.`

ikmsk10: Actual, the file may be any type because it will be used as the binary file without detecting the kind of FS it contains.

commands/bmc.vegman
ikmsk10: Why? The `remote_point` may contain only the IP-address without a specified port.
If `address` was not specified, only this leads to failure.

ikmsk10: Additionally, if we add here checking the remote_point, 
this will lead to a double-coding of failure message declaring.

commands/bmc.vegman
ikmsk10: Fixed

commands/bmc.vegman
ikmsk10: Fixed

commands/host.vegman
alatarum: But, I believe, -e includes -f, so it should be enough to keep only -e

commands/bmc.vegman
alatarum: what happens in case used omit second parameter? 
As far as I see, all of variables will be empty, isn't?
    local remote_point=""
    local address=""
    local port=""
BTW, same for nbd_device - you don't ensure if it set.

commands/bmc.vegman
```diff
@@ -353,3 +353,88 @@ function cmd_reboot {
...
+          return 1
+        fi
+        nbd_device="/dev/nbd${2}"
+        shift
+        ;;
+      *) abort_badarg "${1}";;
+    esac
+    shift
+  done
+
```
alatarum, line 434: Here also: test for nbd_device or set default value

commands/bmc.vegman
ikmsk10: Yes, it might be found as missing, and all the above variables will be empty.
In that case, the port will be filled from the default value (10809),
and the address variable state empty.
The script goes fail and prints message error at line #401.

The `--nbd` second parameter will be checked at the retrieving loop, and if it is omitted,
then all other parameters will be shifted.
And the result of that will be an error of specifying `remote server address.`

commands/bmc.vegman
ikmsk10: The `--nbd NUMBER` parameter is required; it can't be a default.
The check is already defined, but I agree that in case 
`remoteimage disconnect --nbd --yes,` the second parameter `--yes` 
will be dropped and it will be passed as the NUMBER.

I am looking for the right solution, and I am a little confused. Can you give me some ideas?

commands/bmc.vegman
alatarum: ok, I agree about remote_point
but is it allowed to omit "${nbd_device}" parameter for nbd-client?

commands/bmc.vegman
alatarum: Well, saying about the case of "remoteimage disconnect --nbd --yes" it is really tricky to solve. The simplest solution I know is to test if first symbol of ${2} is not '-'. I think it enough for your situation.
But I told about the case of "remoteimage disconnect --yes" - then you will call nbd-client -d "" which I believe not expected.

commands/bmc.vegman
ikmsk10: Well, the `${nbd_device}` won't be empty; check is performed in line #424


```
--nbd)
        if [[ $# -lt 2 ]]; then
          echo "The NBD device number shall be specified" >&2      
```

And the same check implemented for the cmd_remoteimage_disconnect

commands/host.vegman
ikmsk10: Ok, I will suppress the -f. 
Added checking of that isn't a directory

commands/bmc.vegman
ikmsk10: Added test that the first symbol not equal to '-'.

commands/host.vegman
AlexanderAmelkin: @alatarum, no, `-e` doesn't imply `-f`. The `-e` is true for symlinks, directories and device nodes, it just checks existence. The `-f` checks specifically if it's a regular file, it is false for symlinks, dirs and device nodes.

commands/bmc.vegman
AlexanderAmelkin: Leave it as is so far

commands/host.vegman
AlexanderAmelkin: @ikmsk10 your new version with `! -e || -d` looks good.

commands/host.vegman
```diff
@@ -63,6 +63,71 @@ function cmd_power_status {
...
+  subcommand "$@"
+}
+
+# @sudo cmd_virtualmedia_mount admin
+# @doc cmd_virtualmedia_mount
+# Mount a virtual media
+#  --usb - the USB interface type
+#  --usb-ro - the USB interface type (read-only)
+#  --hdd - the HDD interface type
+#  --cdrom - the CDROM interface type (read-only)
```
AlexanderAmelkin: I guess this must go as `the CD/DVD/BD-ROM interface type (read-only), based on file size`. Will that fit?

AlexanderAmelkin (changes requested):

commands/host.vegman
ikmsk10: Yes, this more accurately reflects the functionality of the "--cdrom" flag.

ikmsk10: Fixed.

AlexanderAmelkin (approved):

PR #18, opened 2021-02-03
user: multiple deletion support

nest1ing:
According to design proposal the user deletion call should ask for
confirmation and should accept multiple user names in one call.

This commit brings that.

### DIFF
--- a/commands/user
+++ b/commands/user
@@ -22,6 +22,8 @@ DBUS_ROOT_PATH="/xyz/openbmc_project/user"
 DBUS_ATTRIBUTES="xyz.openbmc_project.User.Attributes"
 DBUS_DELETE="xyz.openbmc_project.Object.Delete"
 
+USERNAME_REGEX="^[\.a-zA-Z0-9_-]+$"
+
 # Lookup for group by role name (privilege in BMC terms)
 function role_to_group {
   case "$1" in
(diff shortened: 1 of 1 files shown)


commands/user
```diff
@@ -120,20 +122,49 @@ function cmd_list {
...
+#   -y, --yes       Do not ask for confirmation
+#   USERNAME...     User account name (multiple names can be added).
 function cmd_delete {
-  if [[ $# -eq 0 ]] || [[ -z "$1" ]]; then
+  local yes=""
+  local names=()
+
+  while [[ $# -gt 0 ]]; do
+    case "$1" in
+      -y | --yes) yes="y";;
```
artemsen, line 134: What if there is a user with name '--yes'? (and his brother 'DROP TABLE' =)

artemsen (approved):

commands/user
nest1ing: These are not allowed user names. It's denied to create with our CLI and I think we shouldn't handle them.

nest1ing: I would say that it will mean you wrote a bad regexp in 0eec5fe0354e22705b98a1f5bf933b228ca3bebe. ;)
But it seems to me that regexp looks fine.

alatarum (approved):

AlexanderAmelkin (approved):

PR #19, opened 2021-02-03
remotemedia: Add the NBD export name flag

ikmsk10:
The many NBD-server implementations require the export name to select
the relevant exporting file.
Additionally, this patch adds a test that the `nbd_number` has been
specified, which there is an important required parameter.

End-user-impact: Now, the `bmc remoteimage` command has a way to specify
                 the NBD exporting name.

Change-Id: I86974aae68b949d025ef874d82a167b96794c714

### DIFF
--- a/commands/bmc.vegman
+++ b/commands/bmc.vegman
@@ -363,22 +363,32 @@ function cmd_remoteimage {
 # @sudo cmd_remoteimage_connect admin
 # @doc cmd_remoteimage_connect
 # Connect to the remote server via the specified protocol
-#   --nbd <0..10>     use NBD protocol to connect the remote server with specified NBD device number
-#   address[:port]    remote server address
+#   --nbd <0..10>         use NBD protocol to connect the remote server with specified NBD device number
+#   -e,--export <name>    the name of the export
+#   address[:port]        remote server address
 function cmd_remoteimage_connect {
     local nbd_device=""
     local remote_point=""
     local address=""
     local port=""
+    local export_name=""
 
     while [[ $# -gt 0 ]]; do
       case "${1}" in
         --nbd)
           if [[ $# -lt 2 ]]; then
-            echo "The NBD device number shall be specified" >&2
+            echo "The ordinal number of the NBD device is required." >&2
             return 1
           fi
           nbd_device="/dev/nbd${2}"
+          shift
+          ;;
+        -e | --export)
+          if [[ $# -lt 2 ]]; then
+            echo "The export name of the NBD device is required" >&2
+            return 1
+          fi
+          export_name="--name ${2} "
           shift
           ;;
         *)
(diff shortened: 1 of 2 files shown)


commands/bmc.vegman
```diff
@@ -363,24 +363,34 @@ function cmd_remoteimage {
 # @sudo cmd_remoteimage_connect admin
 # @doc cmd_remoteimage_connect
 # Connect to the remote server via the specified protocol
-#   --nbd <0..10>     use NBD protocol to connect the remote server with specified NBD device number
-#   address[:port]    remote server address
+#   --nbd <0..10>         use NBD protocol to connect the remote server with specified NBD device number
+#   -e,--export name      the export name from the NBD server
```
alatarum: What is "export name"? Can the comment be a bit more detailed?

ikmsk10: The "export name" is an idiom of NBD server specific. This option means the literal short name of exporting the remote file, which is clear for the NBD-server consumers. 
I don't think that we should provide detailed information about each NBD argument, which might be got from 'man nbd-server'

ikmsk10: But I believe the exported name should be an optional argument.
I will remove the test that this option is present.

artemsen (dismissed):

commands/bmc.vegman
ikmsk10: From nbd-client man:
```
-N | --name -- Specifies the name of the export that we want to use.
 If not specified, nbd-client will ask for a "default" export, if one exists on the server.
```
Well, I can't found any additions to the comment to make it more detailed.

commands/bmc.vegman
```diff
@@ -363,24 +363,34 @@ function cmd_remoteimage {
 # @sudo cmd_remoteimage_connect admin
 # @doc cmd_remoteimage_connect
 # Connect to the remote server via the specified protocol
-#   --nbd <0..10>     use NBD protocol to connect the remote server with specified NBD device number
-#   address[:port]    remote server address
+#   --nbd <0..10>         use NBD protocol to connect the remote server with specified NBD device number
+#   -e,--export <name>    the export name from the NBD server
```
AlexanderAmelkin: Just call it as they do in NBD, "name of the export", not "export name".

commands/bmc.vegman
```diff
@@ -363,24 +363,34 @@ function cmd_remoteimage {
...
-            echo "The NBD device number shall be specified" >&2
+            echo "The ordinal number of the NBD device is required." >&2
             return 1
           fi
           nbd_device="/dev/nbd${2}"
           shift
           ;;
+        -e | --export)
+          if [[ $# -lt 2 ]]; then
+            echo "The export name of the NBD device required" >&2
```
AlexanderAmelkin: `is required`

AlexanderAmelkin (changes requested):

commands/bmc.vegman
ikmsk10: Done

commands/bmc.vegman
ikmsk10: Fixed

AlexanderAmelkin (approved):

PR #20, opened 2021-02-16
health: logs: Use dreport instead of journalctl

nest1ing:
All our platforms have `dreport` from `phosphor-debug-collector` on
board and this tool is more useful for collecting data about the system.

This commit replaces `journalctl` calls with `dreport` ones.

### DIFF
--- a/commands/health.vegman
+++ b/commands/health.vegman
@@ -24,12 +24,36 @@ function cmd_logs {
   subcommand "$@"
 }
 
+# @doc cmd_logs_list
+# Get list of available logs
+function cmd_logs_list {
+    expect_noarg "$@"
+
+    local brief=
+    local name=
+    for entry in /usr/share/dreport.d/pl_user.d/*; do
+        brief="$(awk -F: '/^#\s+@brief:/{print $2}' ${entry})"
+        name="$(realpath ${entry})"
+        printf "%-20s%s\n" "${name##*/}" "${brief}"
+    done
+}
+
 # @sudo cmd_logs_show admin,operator
 # @doc cmd_logs_show
 # Get system logs
 function cmd_logs_show {
-  expect_noarg "$@"
-  journalctl --no-hostname --no-pager
+  local name="${1:-}"
+  local file="/usr/share/dreport.d/plugins.d/${name}"
+
+  for link in /usr/share/dreport.d/pl_user.d/E*${name}; do
+    if [[ -f "${file}" ]] && [[ "$(realpath ${link})" = "${file}" ]]; then
+      DREPORT_INCLUDE="/usr/share/cli/.include" \
+      TIME_STAMP="$(date -u)" exec "${file}"
+      return 1
+    fi
+  done
+
+  abort_badarg "${name}"
 }
 
 # @sudo cmd_logs_export admin,operator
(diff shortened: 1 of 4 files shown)


commands/health.vegman
```diff
@@ -24,22 +24,52 @@ function cmd_logs {
...
 }
 
+# @doc cmd_logs_list
+# Get list of available logs
+function cmd_logs_list {
+    expect_noarg "$@"
+
+    local brief=
+    local name=
+    for entry in /usr/share/dreport.d/pl_user.d/*; do
```
artemsen, line 34: Does Vegman have dreport?

nest1ing: I will bring it there

functions
```diff
@@ -0,0 +1,75 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# @author: Alexander Soldatov <a.soldatov@yadro.com>
+
+# @brief output command result to stdout
+# @param command
+function add_cmd_output()
+{
+    eval $1
```
artemsen: I know that it is a port from Vesnin, but I don't like this `eval`.
It will be executed with root rights, but we cannot control either the commands or their parameters.

nest1ing: It has even come not from Vesnin but from the [upstream](https://github.com/openbmc/phosphor-debug-collector/blob/4c63ce5e2e507a0642e25d3e769ee8cb8638fbfe/tools/dreport.d/include.d/functions#L8). :)
I don't know how do it better. If you have any ideas, please let me know.

Currently there are no any known ways to exploit that bad code.
So, I propose to leave it as is until we found a better way.

alatarum (dismissed):

functions
artemsen: I would prefer not to hack dreport at all =)

artemsen (dismissed):

functions
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env bash
+
+# @brief dreport plugins functions.
+#
+# This file is part of klish project.
```
AlexanderAmelkin: I don't think this file is a parth of klish project anymore after porting it into obmc-yadro-cli.

functions
```diff
@@ -0,0 +1,75 @@
...
+#
+# Copyright (c) 2018 YADRO
```
AlexanderAmelkin: This could be updated

functions
```diff
@@ -0,0 +1,75 @@
...
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
```
AlexanderAmelkin: And this could be replaced with SPDX

AlexanderAmelkin (changes requested):

functions
nest1ing: fixed

functions
nest1ing: done

functions
nest1ing: done

nest1ing:
I found a bug:
```shell
vegman-sx20-QEMU-BMC:~$ health logs show osrelease
/usr/share/dreport.d/plugins.d/osrelease: line 12: add_copy_sym_link_file: command not found
```
Do not merge it until I fix it.

nest1ing:
> I found a bug:
> 
> ```shell
> vegman-sx20-QEMU-BMC:~$ health logs show osrelease
> /usr/share/dreport.d/plugins.d/osrelease: line 12: add_copy_sym_link_file: command not found
> ```
> 
> Do not merge it until I fix it.

Fixed

AlexanderAmelkin:
> Do not merge it until I fix it.

There is WIP check for that. Just put "WIP" in the PR name and merging will be blocked.

AlexanderAmelkin (approved):

PR #23, opened 2021-03-04
Bugfix/bbmc 366 wrong help

AlexanderAmelkin:
Requires YADRO-KNS/obmc-yadro-netconfig#9

For vegman, remove the local help from the `bmc datetime ntpconfig`
function description block and implement calling `netconfig` for help.
For nicole, update the local help to be up to date with `netconfig`.

For all targets pass the CLI command name as argv[0] to
`netconfig` and call it with the new `--cli` option so that
`netconfig` displays the CLI command instead of its own name.

Don't call `fwupdate` twice if both host and bmc versions
are requested. Use a single call with a dynamic grep pattern.

Fix command description for OpenPOWER hosts (nicole, vesnin)

### DIFF
--- a/commands/bmc.nicole
+++ b/commands/bmc.nicole
@@ -180,16 +180,22 @@ function cmd_datetime_show {
   format_date "${compact}" "${utc}" "$(date +%s)"
 }
 
+function call_netconfig() {
+  local FNAME=${FUNCNAME[1]##cmd_}
+  local PNAME="bmc ${FNAME//_/ }"
+  exec -a "${PNAME}" netconfig --cli "$@"
+}
+
 # @sudo cmd_datetime_ntpconfig admin
 # @restrict cmd_datetime_ntpconfig admin
 # @doc cmd_datetime_ntpconfig
 # Configure NTP server connection
-#  {add|del} IP
+#  {add|del} ADDR [ADDR..]
 function cmd_datetime_ntpconfig {
-  if [[ $# -ne 2 ]]; then
-    abort_badarg "Expected exactly 2 arguments"
+  if [[ $# -lt 2 ]]; then
+    abort_badarg "Expected 2 or more arguments"
   fi
-  netconfig ntp "eth0" "$1" "$2"
+  call_netconfig ntp "eth0" "$@"
 }
 
 # @sudo cmd_ifconfig admin
@@ -197,7 +203,7 @@ function cmd_datetime_ntpconfig {
 # @doc cmd_ifconfig
 # BMC network configuration
 function cmd_ifconfig {
-  netconfig "$@"
+  call_netconfig "$@"
 }
 
 # @restrict cmd_syslog admin
--- a/commands/bmc.nicole
+++ b/commands/bmc.nicole
@@ -71,30 +71,28 @@ function cmd_info_uptime {
 # @doc cmd_info_version
 # Show version of BMC and BIOS/UEFI
 #   -b, --bmc   Show BMC version
-#   -u, --bios  Show BIOS/UEFI version
+#   -o, --opfw  Show OpenPOWER Firmware version
 function cmd_info_version {
-  local bmc=""
-  local bios=""
+  local bmc_match='^BMC '
+  local host_match='^Host '
+  local no_match='no-match-pattern'
+  local bmc=${no_match}
+  local host=${no_match}
   while [[ $# -gt 0 ]]; do
     case "$1" in
-      -b | --bmc) bmc="y";;
-      -u | --bios) bios="y";;
+      -b | --bmc) bmc=${bmc_match};;
+      -o | --opfw) host=${host_match};;
       *) abort_badarg "$1";;
     esac
     shift
   done
 
-  if [[ -z "${bmc}${bios}" ]]; then
-    bmc="y"
-    bios="y"
+  if [[ -z "${bmc}${host}" ]]; then
+    bmc=${bmc_match}
+    host=${host_match}
   fi
 
-  if [[ -n "${bmc}" ]]; then
-    fwupdate --version | grep "^BMC "
-  fi
-  if [[ -n "${bios}" ]]; then
-    fwupdate --version | grep "^Host "
-  fi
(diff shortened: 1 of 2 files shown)


commands/bmc.nicole
```diff
@@ -182,24 +180,30 @@ function cmd_datetime_show {
   format_date "${compact}" "${utc}" "$(date +%s)"
 }
 
+function call_netconfig() {
+  local FNAME=${FUNCNAME[1]##cmd_}
```
nest1ing, line 184: looks like unusable variable.

commands/bmc.nicole
```diff
@@ -182,24 +180,30 @@ function cmd_datetime_show {
...
-#  {add|del} IP
+#  {add|del} ADDR [ADDR..]
 function cmd_datetime_ntpconfig {
-  if [[ $# -ne 2 ]]; then
-    abort_badarg "Expected exactly 2 arguments"
+  if [[ $# -lt 2 ]]; then
+    abort_badarg "Expected 2 or more arguments"
   fi
-  netconfig ntp "eth0" "$1" "$2"
+  call_netconfig ntp "eth0" "$1" "$2"
```
nest1ing: I think here is an error.
I guess `call_netconfig ntp eth0 $@` should be here if we want pass several addresses.

commands/bmc.vegman
```diff
@@ -182,24 +180,33 @@ function cmd_datetime_show {
...
-    abort_badarg "Expected exactly 3 arguments"
+  if [[ $# -ge 1 && $1 =~ ^-*h(elp)?$ ]]; then
+    call_netconfig help ntp
+    return
+  fi
+  if [[ $# -lt 3 ]]; then
+    abort_badarg "Expected 3 or more arguments"
   fi
-  netconfig ntp "$1" "$2" "$3"
+  call_netconfig ntp "$1" "$2" "$3"
```
nest1ing: And here too.

commands/bmc.nicole
AlexanderAmelkin: Why? It is used right in the next line. Here we strip `cmd_` from the start of the function name, in the next line we convert underscores into spaces.

commands/bmc.nicole
AlexanderAmelkin: Good catch! (Although that should have been fixed by you when you added this functionality to netconfig ;) )

commands/bmc.vegman
AlexanderAmelkin: Yep.

commands/bmc.nicole
nest1ing: Yep, exactly. I've missed it.

commands/bmc.nicole
```diff
@@ -182,24 +180,30 @@ function cmd_datetime_show {
   format_date "${compact}" "${utc}" "$(date +%s)"
 }
 
+function call_netconfig() {
+  local FNAME=${FUNCNAME[1]##cmd_}
+  local PNAME="bmc ${FNAME//_/ }"
+  exec -a "${PNAME}" netconfig --cli "$@"
```
nest1ing, line 186: `$@` should be without quotes. It is a list of arguments. In quotes it will interpret as a one string with spaces inside.

commands/bmc.nicole
```diff
@@ -182,24 +180,30 @@ function cmd_datetime_show {
...
-#  {add|del} IP
+#  {add|del} ADDR [ADDR..]
 function cmd_datetime_ntpconfig {
-  if [[ $# -ne 2 ]]; then
-    abort_badarg "Expected exactly 2 arguments"
+  if [[ $# -lt 2 ]]; then
+    abort_badarg "Expected 2 or more arguments"
   fi
-  netconfig ntp "eth0" "$1" "$2"
+  call_netconfig ntp "eth0" "$@"
```
nest1ing, line 198: Here is `$@` in quotes again.

commands/bmc.nicole
```diff
@@ -182,24 +180,30 @@ function cmd_datetime_show {
...
 }
 
 # @sudo cmd_ifconfig admin
 # @restrict cmd_ifconfig admin
 # @doc cmd_ifconfig
 # BMC network configuration
 function cmd_ifconfig {
-  netconfig "$@"
+  call_netconfig "$@"
```
nest1ing, line 206: It seems to me here the quotes is also excess.

commands/bmc.nicole
nest1ing: I have a task about this, but I haven't started to solve it yet.

commands/bmc.nicole
AlexanderAmelkin: You'd be right if it was `$*`. However `$@` in double quotes prevents split-and-glob on individual arguments. It translates into `"$1" "$2" "$3" ...` whereas `$@` without quotes translates into `$1 $2 $3` and any argument with spaces or globs in it will be split or substituted. We don't want that.

Here is the example:

```
vegman-sx20-2007210001:~$ function some_func() { for a in $@; do echo $a; done }
vegman-sx20-2007210001:~$ some_func abc "def ghi" jkl mno "pqr stu"
abc
def
ghi
jkl
mno
pqr
stu
vegman-sx20-2007210001:~$ function some_func() { for a in "$@"; do echo $a; done }
vegman-sx20-2007210001:~$ some_func abc "def ghi" jkl mno "pqr stu"
abc
def ghi
jkl
mno
pqr stu
vegman-sx20-2007210001:~$ function some_func() { for a in "$*"; do echo $a; done }
vegman-sx20-2007210001:~$ some_func abc "def ghi" jkl mno "pqr stu"
abc def ghi jkl mno pqr stu
```

commands/bmc.nicole
AlexanderAmelkin: Same reason

commands/bmc.nicole
nest1ing: Yes, you are right

nest1ing (approved):

PR #24, opened 2021-03-15
show error when no groups match regexp

alatarum:
In clicmd when looking up role it is possible no groups match regexp and
GROUP variable left empty. This case each CLI script fails with error:
   line 53: $1: unbound variable
Test if GROUP before use it.

End-user-impact: Show error message like
    Can't find group for user <USERNAME>
Signed-off-by: Andrei Kartashev <a.kartashev@yadro.com>

### DIFF
--- a/clicmd
+++ b/clicmd
@@ -31,15 +31,20 @@ if [[ "${USER}" != "root" || ! -v GROUP ]]; then
   fi
   GROUP=$(id -Gn | sed -n 's/.*\(%GROUP_REGEX%\)[ ]*.*$/\1/gp')
 fi
+if [[ -z "${GROUP}" ]]; then
+  echo "Can't find privilege group for user ${USER}" >&2
+  exit 2
+fi
 
 # Lookup for group by role name (privilege in BMC terms)
 function role_to_group {
-  case "$1" in
+  local role="${1:-}"
+  case "${role}" in
     ${ROLE_ADMIN})    echo "${GROUP_ADMIN}";;
     ${ROLE_OPERATOR}) echo "${GROUP_OPERATOR}";;
     ${ROLE_USER})     echo "${GROUP_USER}";;
     *)
-      echo "Invalid role name: $1" >&2
+      echo "Invalid role name: ${role}" >&2
       return 1
       ;;
   esac
@@ -47,18 +52,19 @@ function role_to_group {
 
 # Lookup for role (privilege in BMC terms) by group name
 function group_to_role {
-  case "$1" in
+  local grp="${1:-}"
+  case "${grp}" in
     ${GROUP_ADMIN})    echo "${ROLE_ADMIN}";;
     ${GROUP_OPERATOR}) echo "${ROLE_OPERATOR}";;
     ${GROUP_USER})     echo "${ROLE_USER}";;
     *)
-      echo "Invalid group name: $1" >&2
+      echo "Invalid group name: ${grp}" >&2
       return 1
       ;;
   esac
 }
 
-ROLE=$(group_to_role ${GROUP})
+ROLE=$(group_to_role "${GROUP}")
 
 # command to execute
 USER_CMD="${BASH_SOURCE[0]}"


clicmd
```diff
@@ -58,6 +58,10 @@ function group_to_role {
   esac
 }
 
+if [[ -z ${GROUP} ]]; then
```
artemsen: Double quotes are mandatory for -z/-n.

nest1ing: It seems to me that the more correct way to solve the issue is fixing the line 50 to something like:
```
case "${1:-}"
```
and the error message at the line 55 to something like:
```
echo "Unable to determine the user's role"
```

What do you think?

alatarum: good idea for 'case "${1:-}"', but IMHO we still should test for -z explicitly.

artemsen (dismissed):

clicmd
nest1ing: What for? 
I mean what is the difference between wrong non empty and just empty values in this case?

alatarum: Just to keep code clear to read

clicmd
```diff
@@ -31,10 +31,14 @@ if [[ "${USER}" != "root" || ! -v GROUP ]]; then
   fi
   GROUP=$(id -Gn | sed -n 's/.*\(%GROUP_REGEX%\)[ ]*.*$/\1/gp')
 fi
+if [[ -z "${GROUP}" ]]; then
+  echo "Can't find group for user ${USER}" >&2
```
nest1ing: It seems to me still incorrect.
We didn't find the group that could help to determine the user's role.
But your message seems to me says the user has no any group's membership at all.

Am I nag? ;)

clicmd
```diff
@@ -47,7 +51,7 @@ function role_to_group {
 
 # Lookup for role (privilege in BMC terms) by group name
 function group_to_role {
-  case "$1" in
+  case "${1:-}" in
```
nest1ing: Empty `${1}` still gives error about unbound variable at line 59.
You should write an error message without `${1}` or use `${1:-}` there too.

clicmd
```diff
@@ -31,10 +31,14 @@ if [[ "${USER}" != "root" || ! -v GROUP ]]; then
...
 fi
+if [[ -z "${GROUP}" ]]; then
+  echo "Can't find group for user ${USER}" >&2
+  exit 2
+fi
 
 # Lookup for group by role name (privilege in BMC terms)
 function role_to_group {
-  case "$1" in
+  case "${1:-}" in
```
nest1ing: The same case as in the `group_to_role`.
The error message at line 46  should be fixed.

nest1ing (changes requested):
The errors  about unbound variable are still possible.

clicmd
alatarum: well.. ok =)

clicmd
```diff
@@ -31,34 +31,40 @@ if [[ "${USER}" != "root" || ! -v GROUP ]]; then
...
 function group_to_role {
-  case "$1" in
+  local grp="${1:-}"
+  case "${grp}" in
     ${GROUP_ADMIN})    echo "${ROLE_ADMIN}";;
     ${GROUP_OPERATOR}) echo "${ROLE_OPERATOR}";;
     ${GROUP_USER})     echo "${ROLE_USER}";;
     *)
-      echo "Invalid group name: $1" >&2
+      echo "Invalid group name: ${grp}" >&2
```
nest1ing, line 61: And now we reached the state when the empty value is also 'Invalid privilege group' and the separate check for empty and non-empty values has no sense. ;)

nest1ing (approved):

clicmd
AlexanderAmelkin: Empty value is checked in line 34, there is no way it can be empty here, or is there any?

nest1ing: Initially I proposed not to do the check at line 34 and use for both cases this function only.

AlexanderAmelkin (approved):

PR #25, opened 2021-03-16
Bugfix/restrict top level help

AlexanderAmelkin:
* help: Restrict top-level commands
* Update permissions per design document changes
* vegman: power: Add `reboot`, fix `off`
* health logs show: Add help

### DIFF
--- a/commands/host.vegman
+++ b/commands/host.vegman
@@ -101,18 +101,34 @@ function cmd_power_off {
     ipmitool chassis power off
     echo "Chassis power is switched off, host is down."
   else
-    ipmitool power off
+    ipmitool power soft
     echo "Shutdown signal sent to host, waiting to power the system off."
   fi
 }
 
-## @sudo cmd_power_reboot admin,operator
-## @restrict cmd_power_status admin,operator
-## @doc cmd_power_reboot
-## Performs host reboot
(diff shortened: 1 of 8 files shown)


commands/host.vegman
```diff
@@ -56,18 +101,35 @@ function cmd_power_off {
...
-#function cmd_power_reboot {
-#  not implemented
-#}
+# @sudo cmd_power_reboot admin,operator
+# @restrict cmd_power_reboot admin,operator
+# @doc cmd_power_reboot
+# Performs host reboot
+#   -y, --yes  Do not ask for confirmation
+function cmd_power_reboot {
+  local force=""
```
nest1ing: will the warm reboot be implemented later or did you just forget it?

AlexanderAmelkin: There is no standard way to implement warm reboot using ipmitool in particular or IPMI in general. There is no such command in the specification. That's why I removed the checks for `force` when copy-pasted the `cmd_power_off` function, but apparently I forgot to remove the `force` variable itself. Will remove it.

nest1ing (approved):

artemsen (approved):

alatarum (approved):

PR #26, opened 2021-03-22
bmc: Various fixes

AlexanderAmelkin:
Make `bmc info version` without arguments
show all firmware versions like it did before.
That was broken since f2ff3222ada5f627a322e920e616c978

End-user-impact: `bmc info version` w/o arguments now
                shows all FW versions again;
                `bmc datetime show` command now has correct help;
                Fixed a typo in help message of `bmc syslog reset` command

### DIFF
--- a/commands/bmc.nicole
+++ b/commands/bmc.nicole
@@ -73,11 +73,15 @@ function cmd_info_uptime {
 #   -b, --bmc   Show BMC version
 #   -o, --opfw  Show OpenPOWER Firmware version
 function cmd_info_version {
+  # If no options are given, show everything
+  [[ $# -eq 0 ]] && set -- -b -o
+
   local bmc_match='^BMC '
   local host_match='^Host '
   local no_match='no-match-pattern'
   local bmc=${no_match}
   local host=${no_match}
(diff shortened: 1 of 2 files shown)


artemsen (dismissed):

commands/bmc.nicole
```diff
@@ -73,11 +73,15 @@ function cmd_info_uptime {
...
 function cmd_info_version {
+  # If no options are given, show everything
+  [[ $# -eq 0 ]] && set -- -b -o
+
   local bmc_match='^BMC '
   local host_match='^Host '
   local no_match='no-match-pattern'
   local bmc=${no_match}
   local host=${no_match}
+
```
nest1ing, line 84: This line does not look like something improves the readable but it also does not look bad.
I think you should either remove it or add the same line into the `bmc.vegman`.

nest1ing (dismissed):

commands/bmc.nicole
AlexanderAmelkin: Will do

nest1ing (dismissed):

PR #30, opened 2021-03-30
Shellcheck and address what it finds

AlexanderAmelkin:
Shellcheck finds a lot of problems in this project.
Add a github workflow to automatically run it,
and fix everything that it finds.

### DIFF
--- a/.github/workflows/blank.yml
+++ b/.github/workflows/blank.yml
@@ -1,0 +1,29 @@
+# This is a basic workflow to help you get started with Actions
+
+name: CI
+
+# Controls when the action will run. 
+on:
+  # Triggers the workflow on push or pull request events but only for the master branch
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "check"
+  check:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - uses: actions/checkout@v2
+
+      - name: ShellCheck
+        uses: ludeeus/action-shellcheck@1.0.0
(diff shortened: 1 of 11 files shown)


.github/workflows/blank.yml
```diff
@@ -0,0 +1,29 @@
+# This is a basic workflow to help you get started with Actions
+
+name: CI
+
+# Controls when the action will run. 
+on:
+  # Triggers the workflow on push or pull request events but only for the master branch
+  push:
+    branches: [ master ]
```
artemsen, line 9: I wouldn't restrict the action with master branch only.
This will provide an opportunity to fix problems even before the PR is created.

AlexanderAmelkin: How would one know then that there is a problem?

artemsen: I mean, let's run shellcheck on any branch, not just master. 
With current settings, we have to create "PR to master" to start checking.

commands/bmc.vegman
```diff
@@ -152,14 +152,15 @@ function cmd_datetime_method {
...
+  # Only check format here, values are checked for sanity by `date`
+  if ! [[ ${1:-} =~ ^[0-9]{4}(-[0-9]{2}){5}$ ]]; then
     echo "Please enter date and time in YYYY-MM-DD-HH-mm-SS format" >&2
     return 1
   fi
 
   local new_date="${1:0:10}"
   local new_time="${1:11}"
-  date "${new_date} ${new_time//-/:}"
+  echo date "${new_date}${new_time//-/:}"
```
artemsen: echo?

.github/workflows/blank.yml
AlexanderAmelkin: I get it, but if we run shellcheck on any branch, how would we know that it failed? By going into the Actions page and searching for our new branch? Why not via a PR then?

commands/bmc.vegman
AlexanderAmelkin: Great catch! It's a leftover from my debugging of the proper regexp. Will remove it!

.github/workflows/blank.yml
artemsen: 1. I make changes.
2. I create PR.
3. I see an error in check, I fix it, I update PR
4. goto 3
So, user have to create PR to check his work. But may be you are right. User can install shellcheck locally.

image
Green marker tell us that we have passed the check.

artemsen (approved):

nest1ing (approved):

alatarum (approved):

PR #31, opened 2021-04-01
health: logs: fix error on show hostlogger

nest1ing:
The command `health logs show ...` has an incorrect TIME_STAMP definition.
It leads to the error when command tries to log its own messages.
    
For example: the command `health logs show hostlogger` fails when there are no hostlogger files.
    
This commit fixes the issue.

### DIFF
--- a/commands/health.nicole
+++ b/commands/health.nicole
@@ -59,7 +59,7 @@ function cmd_logs_show {
     if [[ -f "${file}" ]] && [[ "$(realpath "${link}")" = "${file}" ]]; then
       # Should exec fail, the script will abort due to `set -e` in effect
       DREPORT_INCLUDE="/usr/share/cli/.include" \
-      TIME_STAMP=$(date -u) exec "${file}"
+      TIME_STAMP="date -u" exec "${file}"
     fi
   done
 
--- a/commands/health.vegman
+++ b/commands/health.vegman
@@ -60,7 +60,7 @@ function cmd_logs_show {
     if [[ -f "${file}" ]] && [[ "$(realpath "${link}")" = "${file}" ]]; then
       # Should exec fail, the script will abort due to `set -e` in effect
       DREPORT_INCLUDE="/usr/share/cli/.include" \
-      TIME_STAMP=$(date -u) exec "${file}"
+      TIME_STAMP="date -u" exec "${file}"
     fi
   done
 


alatarum (dismissed):

functions
```diff
@@ -44,15 +44,15 @@ function add_copy_sym_link_file()
 # @param error message
 function log_error()
 {
-   echo "$($TIME_STAMP) ERROR: $@"
+   echo "${TIME_STAMP} ERROR: $@"
```
artemsen: Where is this variable defined?
Originally, it is a `date` call:
https://github.com/openbmc/phosphor-debug-collector/blob/master/tools/dreport.d/dreport#L51

nest1ing: It is defined exactly in the place that you specified. ;)

artemsen: Really? How?
Even so, now instead of timestamp it will print:
`date -t ERROR: blah-blah-blah...`

nest1ing: No, I'm wrong. For the CLI it defined [there](https://github.com/YADRO-KNS/obmc-yadro-cli/blob/cc2cd3f387b292392949ba9c36c5c99935fe5234/commands/health.vegman#L62).

artemsen (dismissed):

functions
nest1ing: And you are right, the root of the issue in other place.
I've rework this PR.

artemsen (dismissed):

artemsen (approved):

commands/health.nicole
```diff
@@ -59,7 +59,7 @@ function cmd_logs_show {
     if [[ -f "${file}" ]] && [[ "$(realpath "${link}")" = "${file}" ]]; then
       # Should exec fail, the script will abort due to `set -e` in effect
       DREPORT_INCLUDE="/usr/share/cli/.include" \
-      TIME_STAMP=$(date -u) exec "${file}"
+      TIME_STAMP="date -u" exec "${file}"
```
AlexanderAmelkin, line 62: The question is: do we actually need any time stamp at all when printing error messages to the user?

nest1ing: This line restores the original `dreport`'s behavior.
If you think that we should change it, we can discuss your proposal and make it as a standalone change.

AlexanderAmelkin (approved):

PR #33, opened 2021-04-26
vegman: add audit logs for user actions

fr0st61te:
Tracing user actions via log_audit.

Signed-off-by: Ivan Mikhaylov <i.mikhaylov@yadro.com>

### DIFF
--- a/clicmd
+++ b/clicmd
@@ -83,6 +83,21 @@ function exec_tool {
   exec -a "${PNAME}" "${TOOL}" "$@"
 }
 
+# @brief log audit message
+# @param message
+function log_audit()
+{
+if [ -n "${4-}" ]; then
+  echo "$1:" "${SUDO_USER:-${USER}}:" "$(sed -e 's/\/usr\/bin\///' <<< "$2")" "$3" "$4" | systemd-cat -t CLI
+  return
+fi
+if [ -n "${3-}" ]; then
+  echo "$1:" "${SUDO_USER:-${USER}}:" "$(sed -e 's/\/usr\/bin\///' <<< "$2")" "$3" | systemd-cat -t CLI
+  return
+fi
+echo "CLI audit log function called with wrong arguments:" "$@" | systemd-cat -t CLI
+}
+
 # Handle subcommand that requires additional arguments.
 function subcommand {
   local parent="${FUNCNAME[1]}"
@@ -125,8 +140,11 @@ function subcommand {
   # if command has @sudo tag - we need to run the function with elevated privileges
   if grep -q "^#\\s*@sudo\\s\\+${cmd}\\s" "${REAL_CMD}" && [[ ${EUID} -ne 0 ]]; then
     local sub_path="${cmd/#cmd_}"
+    local rc=0
     # shellcheck disable=SC2086
-    GROUP=${GROUP} exec sudo "${USER_CMD}" ${sub_path//_/ } "$@"
+    GROUP=${GROUP} sudo "${USER_CMD}" ${sub_path//_/ } "$@" || rc="$?"
+    log_audit "$rc" "${USER_CMD}" "${sub_path//_/ }" "$@"
+    exit "$rc"
   fi
 
   # execute subcommand


functions
```diff
@@ -76,3 +76,10 @@ function log_summary()
     echo "$($TIME_STAMP) " "$@"
 }
 
+# @brief log audit message
+# @param message
+function log_audit()
```
AlexanderAmelkin: Wrong place for this. The `functions` file is a replacement for `phosphor-debug-collector` to work as `health logs` command. If you want to make some function universally available to all `obmc-yadro-cli` scripts, simply add it to `clicmd`.

clicmd
```diff
@@ -74,6 +74,8 @@ if [[ ! -f "${REAL_CMD}" ]]; then
   exit 2
 fi
 
+source /usr/share/cli/.include/functions
```
AlexanderAmelkin: Don't do this. Add `log_audit` here directly.

AlexanderAmelkin (changes requested):

AlexanderAmelkin:
Please also mind the ShellCheck CI findings. ShellCheck must pass.

commands/bmc.vegman
```diff
@@ -345,6 +346,7 @@ function cmd_update_start {
     return 1
   fi
 
+  log_audit "BMC/BIOS update"
   fwupdate ${reset} ${yes} -f "${fwfile}"
```
AlexanderAmelkin: I guess we also need to modify `fwupdate` to log the type of firmware being updated and the versions (from/to).
Mind that `fwupdate` can be called from WebUI as well as from CLI.

We also need to log whether or not the update attempt was successful. Mind that if `fwupdate` fails, the script will be aborted due to `set -e`, and nothing will be executed after this line, so logging of the status must be added either into `fwupdate` itself, or as an `&&` + `||` pair after the `fwupdate` call.

commands/bmc.vegman
```diff
@@ -345,6 +346,7 @@ function cmd_update_start {
     return 1
   fi
 
+  log_audit "BMC/BIOS update"
```
AlexanderAmelkin: It's better to say "Firmware update start" here

commands/bmc.vegman
```diff
@@ -379,6 +381,7 @@ function cmd_reboot {
   fi
 
   echo "Reboot the BMC..."
+  log_audit "BMC reboot"
   reboot ${force}
 }
```
AlexanderAmelkin: Why did you choose to not log the `remoteimage` actions despite that you do log the `virtualmedia` ones?
The `remoteimage` command allows for mounting a remote nbd and then use that file as an argument to `virtualmedia mount`. It will then be unclear what is that nbd share file and where did it come from if you don't log it here.

commands/bmc.vegman
```diff
@@ -304,6 +304,7 @@ function cmd_config_default {
     esac
     shift
   done
+  log_audit "Restore the factory default BMC config"
```
AlexanderAmelkin: Why don't you log the `datetime method`, `datetime set`, `datetime ntpconfig`, `ifconfig`, and `syslog` configuration commands?

For `ifconfig` and `datetime ntpconfig` I think the logging must be done inside `netconfig`.

commands/host.vegman
```diff
@@ -225,6 +230,7 @@ function cmd_virtualmedia_umount {
     confirm
   fi
 
+  log_audit "Host virtual media eject"
   usb-ctrl eject "usb.ms.cli"
 }
```
AlexanderAmelkin: I'm quite sure that we need to log the fact of connection to the host console.

commands/user
```diff
@@ -67,7 +67,7 @@ function cmd_create {
 
   local priv
   priv=$(role_to_group "${role}")
- 
+  log_audit "Create user ${name}, ${role}"
```
AlexanderAmelkin: I busctl fails, the user won't be created but there will be an entry in the audit log.
I think that we need to log only successful commands or better yet log the command execution status afterwards.
This comment relates to all commands, not just this one.

commands/user
```diff
@@ -182,12 +183,14 @@ function cmd_set_password {
...
   fi
   passwd "${name}"
+  log_audit "Update password for ${name}"
 }
 
 # @doc cmd_set_mypassword
 # Change own password
 function cmd_set_mypassword {
   expect_noarg "$@"
+  log_audit "Password update"
```
AlexanderAmelkin: Better say `"Change of own password"`

commands/user
```diff
@@ -182,12 +183,14 @@ function cmd_set_password {
     return 1
   fi
   passwd "${name}"
+  log_audit "Update password for ${name}"
```
AlexanderAmelkin: Better say `"Change of password for ${name}'`

commands/user
```diff
@@ -229,6 +232,7 @@ function cmd_set_role {
   local priv
   priv=$(role_to_group "${role}")
 
+  log_audit "Update user : ${name}, role : ${role}"
```
AlexanderAmelkin: Better say `"Set role ${role} for user ${name}"`.

AlexanderAmelkin (changes requested):
`diagnostics led on/off` and `health logs clear` commands also need to be logged.
Also, when changing one target (e.g. vegman), please don't forget to update the others as well. Let's keep CLI identical in behavior as much as possible across all supported targets. Note that `vesnin` scripts are just symlinks.

commands/bmc.vegman
nest1ing: > I guess we also need to modify `fwupdate` to log the type of firmware being updated and the versions (from/to).
> Mind that `fwupdate` can be called from WebUI as well as from CLI.

There is some possible problems. The version info is available only from the bundle (when MANIFEST is present). But WebUI called fwupdate with a standalone partition image and its version is unknown.

> We also need to log whether or not the update attempt was successful. Mind that if `fwupdate` fails, the script will be aborted due to `set -e`, and nothing will be executed after this line, so logging of the status must be added either into `fwupdate` itself, or as an `&&` + `||` pair after the `fwupdate` call.

Agree.

nest1ing (commented):
All actions to change something always called with `sudo` that already logged each call.
What benefit of this changes?

AlexanderAmelkin:
> All actions to change something always called with `sudo` that already logged each call.
> What benefit of this changes?

Good point. Maybe we just need to add those sudo's log entries into the audit log. @fr0st61te, what do you think?

fr0st61te:
> > All actions to change something always called with `sudo` that already logged each call.
> > What benefit of this changes?
> 
> Good point. Maybe we just need to add those sudo's log entries into the audit log. @fr0st61te, what do you think?

I have to look at pam documentation, maybe some posthooks exists for such behavior.

clicmd
```diff
@@ -74,6 +74,18 @@ if [[ ! -f "${REAL_CMD}" ]]; then
   exit 2
 fi
 
+# @brief log audit message
+# @param message
+function log_audit()
+{
+if [ -n "${4-}" ]; then
```
nest1ing, line 90: It seems to me that the check for the presence of `${1}` and `${2}` is wanted to prevent illegal usage.

nest1ing (approved):

clicmd
```diff
@@ -116,7 +128,9 @@ function subcommand {
   # if command has @sudo tag - we need to run the function with elevated privileges
   if grep -q "^#\\s*@sudo\\s\\+${cmd}\\s" "${REAL_CMD}" && [[ ${EUID} -ne 0 ]]; then
     local sub_path="${cmd/#cmd_}"
-    GROUP=${GROUP} exec sudo "${USER_CMD}" "${sub_path//_/ }" "$@"
+    GROUP=${GROUP} sudo "${USER_CMD}" "${sub_path//_/ }" "$@"
+    log_audit "$?" "${USER_CMD}" "${sub_path//_/ }" "$@"
```
nest1ing: It won't execute when `sudo` fail because `bash -e` is used.
Try to wrap this two lines with `set +e ... set -e`

nest1ing (changes requested):

clicmd
nest1ing: Or even better, use something like:
```bash
    RC=0
    GROUP=${GROUP} sudo "${USER_CMD}" "${sub_path//_/ }" "$@" || RC=$?
    log_audit "${RC}" "${USER_CMD}" "${sub_path//_/ }" "$@"
```
It looks more honest ;)

clicmd
AlexanderAmelkin: The `-` mitigates that and produces an empty string if `${1}`, `${2}` or `${3}` aren't present.

clicmd
```diff
@@ -74,6 +74,24 @@ if [[ ! -f "${REAL_CMD}" ]]; then
...
+if [ -n "${4-}" ]; then
+  echo "$1:" "${SUDO_USER:-${USER}}:" "$(sed -e 's/\/usr\/bin\///' <<< "$2")" "$3" "$4" | systemd-cat -t CLI
+  return
+fi
+if [ -n "${3-}" ]; then
+  echo "$1:" "${SUDO_USER:-${USER}}:" "$(sed -e 's/\/usr\/bin\///' <<< "$2")" "$3" | systemd-cat -t CLI
+  return
+fi
+if [ -n "${2-}" ] || [ -n "${1-}" ]; then
+  echo "Something went wrong with cli access!" | systemd-cat -t CLI
```
AlexanderAmelkin: capitalize CLI please

clicmd
fr0st61te: Yeap, looks like!

AlexanderAmelkin (changes requested):

clicmd
nest1ing: Yes, and the CLI will keep silence if somebody calls this function with the wrong arguments.

nest1ing (dismissed):

clicmd
AlexanderAmelkin: @nest1ing, there is just one caller, and the parameters are always there.

clicmd
```diff
@@ -83,6 +83,23 @@ function exec_tool {
...
+{
+if [ -n "${4-}" ]; then
+  echo "$1:" "${SUDO_USER:-${USER}}:" "$(sed -e 's/\/usr\/bin\///' <<< "$2")" "$3" "$4" | systemd-cat -t CLI
+  return
+fi
+if [ -n "${3-}" ]; then
+  echo "$1:" "${SUDO_USER:-${USER}}:" "$(sed -e 's/\/usr\/bin\///' <<< "$2")" "$3" | systemd-cat -t CLI
+  return
+fi
+if [ -n "${2-}" ] || [ -n "${1-}" ]; then
```
AlexanderAmelkin: You're checking $2 and $1 here for being non-empty, yet don't use them in any way afterwards. Why check then?
Did you mean `-z` instead? If so, then why at the very end?
I can't get what this case is for.

AlexanderAmelkin (approved):

PR #34, opened 2021-06-01
Feature/bbmc

nonameletters:


commands/host.vegman
```diff
@@ -224,6 +224,55 @@ function cmd_virtualmedia_umount {
   usb-ctrl eject "usb.ms.cli"
 }
 
+# @restrict cmd_config tech,admin
```
alatarum: don't mention tech user here, it will be added by recipe later.

commands/host.vegman
```diff
@@ -224,6 +224,55 @@ function cmd_virtualmedia_umount {
...
+      ipmitool chassis bootdev none
+  }>/dev/null
+}
+
+# @sudo cmd_config_status tech,admin
+# @restrict cmd_config_status tech,admin
+# @doc cmd_config_status
+# Show boot settings
+function cmd_config_status {
+    ipmitool chassis bootparam get 5
```
nest1ing: I guess that `expect_noarg "$@"` should be at the first line.

commands/host.vegman
```diff
@@ -224,6 +224,55 @@ function cmd_virtualmedia_umount {
...
 }
 
+# @restrict cmd_config tech,admin
+# @doc cmd_config
+# Host config backup and restore
+function cmd_config {
+  subcommand "$@"
+}
+
+# @sudo cmd_config_default tech,admin
```
nest1ing: Remove `tech` user mentioning as it's for internal using only and will be added by yocto.

commands/host.vegman
```diff
@@ -224,6 +224,55 @@ function cmd_virtualmedia_umount {
...
+# @restrict cmd_config_default tech,admin
+# @doc cmd_config_default
+# Restore the factory default Host config
```
nest1ing: If the confirmation is really required, describe it here

commands/host.vegman
```diff
@@ -224,6 +224,53 @@ function cmd_virtualmedia_umount {
...
+  fi
+  
+  ipmitool chassis bootdev none clear-cmos=yes >/dev/null
+}
+
+# @sudo cmd_config_nodefault tech,admin
+# @restrict cmd_config_nodefault tech,admin
+# @doc cmd_config_nodefault
+# Undo if the factory default Host was set
+function cmd_config_nodefault {
```
alatarum: I would make this function and status as subcommands (options) of cmd_config_default, like:
> host config default --cancel
> host config default --status

commands/host.vegman
```diff
@@ -224,6 +224,53 @@ function cmd_virtualmedia_umount {
...
+  echo "Undo Host BIOS setting reset"
```
nest1ing: I assume the confirmation is also required here.
If not - add `expect_noarg "$@"` at the first line.

commands/host.vegman
```diff
@@ -224,6 +224,53 @@ function cmd_virtualmedia_umount {
...
+  
+  ipmitool chassis bootdev none >/dev/null
```
alatarum: I'm bit confused, why you use confirmation to set flag, but don't use to reset it... But maybe it have sence

nest1ing (changes requested):

commands/host.vegman
```diff
@@ -224,6 +224,53 @@ function cmd_virtualmedia_umount {
...
+}
+
+# @sudo cmd_config_status tech,admin
+# @restrict cmd_config_status tech,admin
+# @doc cmd_config_status
+# Show boot settings
+function cmd_config_status {
+    ipmitool chassis bootparam get 5
```
alatarum: Shouldn't you process the answer? How it looks like?

nonameletters: Boot parameter version: 1
Boot parameter 5 is valid/unlocked
Boot parameter data: 8000000000
 Boot Flags :
   - Boot Flag Valid
   - Options apply to only next boot
   - BIOS PC Compatible (legacy) boot 
   - Boot Device Selector : No override
   - BIOS verbosity : System Default
   - Console Redirection control : Console redirection occurs per BIOS configuration setting (default)
   - BIOS Mux Control Override : BIOS uses recommended setting of the mux at the end of POST

commands/host.vegman
```diff
@@ -224,6 +224,84 @@ function cmd_virtualmedia_umount {
...
+# @sudo cmd_config_default admin,operator,user
+# @doc cmd_config_default
+# Host set/reset UEFI boot configuration on next boot.
+# Possible flags: --cancel, --status
+function cmd_config_default {
+  local count=$#
+
+  while [[ $# -gt 0 ]]; do
+    case "${1}" in
+      --cancel) loc_config_default_cancel;;
```
nest1ing: `--yes` will never be passed to the function.
As a result the confirmation will be asked in any case.

commands/host.vegman
```diff
@@ -224,6 +224,84 @@ function cmd_virtualmedia_umount {
...
+# @doc cmd_config
+# Host configuration
+function cmd_config {
+  subcommand "$@"
+}
+
+# @sudo cmd_config_default admin,operator,user
+# @doc cmd_config_default
+# Host set/reset UEFI boot configuration on next boot.
+# Possible flags: --cancel, --status
```
nest1ing: I assume the options should be explained more detailed.
And `--yes` also should present here.
This comment is used as a help message shown by `host config default --help`.

nest1ing (dismissed):

commands/host.vegman
```diff
@@ -224,6 +224,86 @@ function cmd_virtualmedia_umount {
   usb-ctrl eject "usb.ms.cli"
 }
 
+
+function loc_config_default_set {
```
AlexanderAmelkin: No need to call them `loc_*`. Anything that doesn't start with `cmd_` is safe. These scripts are anyway sourced into `clicmd`.

commands/host.vegman
```diff
@@ -224,6 +224,86 @@ function cmd_virtualmedia_umount {
...
+    case "$1" in
+      -y | --yes) yes="y";;
+      *) abort_badarg "$1";;
+    esac
+    shift
+  done
+
+  echo "Cancel reset to default host settings on next boot."
+  if [[ -z "${yes}" ]]; then
+    confirm
```
AlexanderAmelkin, line 252: I guess we also need to state here undert this if that "This operation will also clear the boot device settings, even if the request to clear settings was not actually in effect".

Or we could use `busctl` below to only reset the clear-cmos option without affecting the other boot settings.

commands/host.vegman
```diff
@@ -224,6 +224,86 @@ function cmd_virtualmedia_umount {
...
+  ipmitool chassis bootdev none >/dev/null
+}
+
+function loc_config_default_status {
+  expect_noarg "$@"
+  strArr=($(ipmitool chassis bootparam get 5|grep -i 'Boot parameter data'))
+  val=${strArr[3]}
+  len=${#val}
+
+  if [[ $((16#${val:$len-8:1})) -ge 8 ]]; then
```
AlexanderAmelkin: Why all this magic? Why not just grep `ipmitool` output for `CMOS Clear` ? As far as I can see, that's what `ipmitool` would pring if the flag is set:

```c
if(rsp->data[3] & BF2_CMOS_CLEAR)                                 
        printf("   - CMOS Clear\n"); 
```

commands/host.vegman
```diff
@@ -224,6 +224,86 @@ function cmd_virtualmedia_umount {
...
+
+# @sudo cmd_config_default admin,operator,user
+# @doc cmd_config_default
+# Host set/reset UEFI boot configuration on next boot.
+# Possible flags: --cancel, --status, --yes to approve without prompt
+function cmd_config_default {
+  if [[ $# -gt 0 ]]; then
+    case "${1}" in
+      --cancel)
+                  if [[ $# -ge 2 ]]; then
```
AlexanderAmelkin: Something is wrong with indentation here. Please do it like this:
```
function...
  if...
    case...
      --cancel)
        cancel_code
        ;;
      --status)
        status_code
        ;;
      -y | --yes)
        yes_code
        ;;
      *) echo ...
...

File: commands/host.vegman [hunk 15]
```diff
@@ -224,6 +224,86 @@ function cmd_virtualmedia_umount {
...
+                    loc_config_default_cancel ${2}
+                  else
+                    loc_config_default_cancel
+                  fi;;
+      --status)   loc_config_default_status;;
+      -y | --yes) loc_config_default_set ${1};;
```
AlexanderAmelkin: Why? Can't I use `--yes` with `--cancel` ? That's strange.
I think you need to move `--yes` processing out of the set and cancel functions.
You could determine the name of the function to call in this option processing loop, and then just call it with just a single `${yes}` argument.

commands/host.vegman
```diff
@@ -224,6 +224,86 @@ function cmd_virtualmedia_umount {
...
+# @doc cmd_config
+# Host configuration
+function cmd_config {
+  subcommand "$@"
+}
+
+# @sudo cmd_config_default admin,operator,user
+# @doc cmd_config_default
+# Host set/reset UEFI boot configuration on next boot.
+# Possible flags: --cancel, --status, --yes to approve without prompt
```
AlexanderAmelkin: This doesn't look like options help in other commands. Take a look, for example, at `cmd_info_version` in `bmc.vegman`:

```
# @doc cmd_info_version
# Show version of BMC and BIOS/UEFI
#   -b, --bmc   Show BMC version
#   -u, --bios  Show BIOS/UEFI version
```

commands/host.vegman
```diff
@@ -224,6 +224,86 @@ function cmd_virtualmedia_umount {
...
+}
+
+# @sudo cmd_config admin,operator,user
+# @doc cmd_config
+# Host configuration
+function cmd_config {
+  subcommand "$@"
+}
+
+# @sudo cmd_config_default admin,operator,user
```
AlexanderAmelkin: A `@restrict` clause is definitely missing. The `@sudo` just tells the script that it must creade a sudoers entry, `@restrict` actually restricts the users from seeing and executing the command.

Also, both `@restrict` and `@sudo` must contain just `admin` for this command.

AlexanderAmelkin (changes requested):

commands/host.vegman
```diff
@@ -224,6 +224,96 @@ function cmd_virtualmedia_umount {
...
+
+# @sudo cmd_config_default admin
+# @restrict cmd_config_default admin
+# @doc cmd_config_default
+# Host set/reset UEFI boot configuration on next boot.
+#  --cancel  - Undo reset UEFI boot configuration, if it was set by mistake
+#  --status  - Shows will be or not UEFI reset during next boot
+#  -y, --yes - does not ask for interactive confirmation
+function cmd_config_default {
+  if [[ $# -gt 0 ]]; then
```
alatarum: I suggest to write something like 
```
  local yes=""
  local action="set"
  while [[ $# -gt 0 ]]; do
    case "$1" in
      --cancel) action="cancel";;
      --status) action="status";;
      -y | --yes) yes="y";;
      *) abort_badarg "$1";;
    esac
    shift
  done
```
and then call function depend on ${action} value

AlexanderAmelkin: Exactly. That's what I suggested earlier.

AlexanderAmelkin: Hmm... I somehow don't see my comment where I suggested exactly that.

commands/host.vegman
```diff
@@ -224,6 +224,96 @@ function cmd_virtualmedia_umount {
...
+  subcommand "$@"
+}
+
+# @sudo cmd_config_default admin
+# @restrict cmd_config_default admin
+# @doc cmd_config_default
+# Host set/reset UEFI boot configuration on next boot.
+#  --cancel  - Undo reset UEFI boot configuration, if it was set by mistake
+#  --status  - Shows will be or not UEFI reset during next boot
+#  -y, --yes - does not ask for interactive confirmation
```
AlexanderAmelkin: "does not" -> "Do not"

commands/host.vegman
AlexanderAmelkin: Please do as Andrey suggests here and I suggested in the previous review round.

AlexanderAmelkin (changes requested):

commands/host.vegman
AlexanderAmelkin: You're still doing magic. Please either explain why, or switch to checking for "CMOS Clear" in the output of `ipmitool`.

commands/host.vegman
```diff
@@ -224,6 +224,90 @@ function cmd_virtualmedia_umount {
...
+
+function config_default_set {
+  local yes=""
+  while [[ $# -gt 0 ]]; do
+    case "$1" in
+      -y | --yes) yes="y";;
+      *) abort_badarg "$1";;
+    esac
+    shift
+  done
```
AlexanderAmelkin: You don't need this anymore. You either pass `-y` in `$1` (that is, a non-empty `$1`), or you don't pass anything (empty `$1`). You only need this:

```
local yes="${1-}"

if [[ -z "${yes}" ]]; then
...
```

That was the whole idea behind determining the action in the main function that you don't have to repeat the code here and in cancel function.

commands/host.vegman
```diff
@@ -224,6 +224,90 @@ function cmd_virtualmedia_umount {
...
+
+function config_default_cancel {
+  local yes=""
+  while [[ $# -gt 0 ]]; do
+    case "$1" in
+      -y | --yes) yes="y";;
+      *) abort_badarg "$1";;
+    esac
+    shift
+  done
```
AlexanderAmelkin: same here

commands/host.vegman
```diff
@@ -224,6 +224,90 @@ function cmd_virtualmedia_umount {
...
+# @doc cmd_config
+# Host configuration
+function cmd_config {
+  subcommand "$@"
+}
+
+# @sudo cmd_config_default admin
+# @restrict cmd_config_default admin
+# @doc cmd_config_default
+# Host set/reset UEFI boot configuration on next boot.
```
AlexanderAmelkin: Change to "Reset UEFI BIOS configuration to defaults on the next boot or cancel that request"

commands/host.vegman
```diff
@@ -205,7 +205,7 @@ function cmd_virtualmedia_mount {
 # @restrict cmd_virtualmedia_umount admin
 # @doc cmd_virtualmedia_umount
 # Unmount the virtual media
-#  -y, --yes  - does not ask for interactive confirmation
+#  -y, --yes  - do not ask for interactive confirmation
```
AlexanderAmelkin: While you're at it, please change it to capitalized 'Do'. Also, please make this refactoring a separate commit (and please check in other files where `--yes` is specified. Let's make the messge identical everywhere).

commands/host.vegman
```diff
@@ -224,6 +224,90 @@ function cmd_virtualmedia_umount {
...
+function cmd_config {
+  subcommand "$@"
+}
+
+# @sudo cmd_config_default admin
+# @restrict cmd_config_default admin
+# @doc cmd_config_default
+# Host set/reset UEFI boot configuration on next boot.
+#  --cancel  - Undo reset UEFI boot configuration, if it was set by mistake
+#  --status  - Shows will be or not UEFI reset during next boot
```
AlexanderAmelkin: Change to "Show the status of the settings reset request"

commands/host.vegman
```diff
@@ -224,6 +224,90 @@ function cmd_virtualmedia_umount {
...
+# Host configuration
+function cmd_config {
+  subcommand "$@"
+}
+
+# @sudo cmd_config_default admin
+# @restrict cmd_config_default admin
+# @doc cmd_config_default
+# Host set/reset UEFI boot configuration on next boot.
+#  --cancel  - Undo reset UEFI boot configuration, if it was set by mistake
```
AlexanderAmelkin: Change to "Cancel the settings reset request if it hasn't been processed yet"

commands/host.vegman
```diff
@@ -224,6 +224,90 @@ function cmd_virtualmedia_umount {
...
+    esac
+    shift
+  done
+
+  case "$action" in
+    cancel) config_default_cancel $yes;;
+    status) config_default_status;;
+    set)    config_default_set $yes;;
+    *) echo "Unsupported parameter. Use --status, --cancel or --yes to
+approve";;
```
AlexanderAmelkin: Please don't indent like this. Close the quote and use line continuation `\` to do proper indentation:

```
echo "Long string starts here" \
     "and continues on the next line" \
     "with proper indentation."
```

AlexanderAmelkin (changes requested):

commands/host.vegman
```diff
@@ -224,6 +224,90 @@ function cmd_virtualmedia_umount {
   usb-ctrl eject "usb.ms.cli"
 }
 
+
+function config_default_set {
+  local yes=""
+  while [[ $# -gt 0 ]]; do
```
nest1ing: I guess that this code is excess now as the `--yes` is already checked by the caller and passed to here only when it really specified by the user.
So, it seems to me, that `[[ -z "${1:-}" ]]` is enough to determine the confirmation is needed.

commands/host.vegman
```diff
@@ -224,6 +224,90 @@ function cmd_virtualmedia_umount {
...
+    esac
+    shift
+  done
+
+  echo "Host BIOS settings will be reset to default on next boot."
+  if [[ -z "${yes}" ]]; then
+    confirm
+  fi
+
+  ipmitool chassis bootdev none clear-cmos=yes >/dev/null
```
nest1ing, line 244: I suggest to use `busctl` here too, as it would be a shorter way.

commands/host.vegman
```diff
@@ -224,6 +224,90 @@ function cmd_virtualmedia_umount {
...
+  fi
+
+  busctl set-property xyz.openbmc_project.Settings \
+         /xyz/openbmc_project/control/host0/boot/one_time \
+         xyz.openbmc_project.Control.Boot.ClearCmos ClearCmos b false
+}
+
+function config_default_status {
+  expect_noarg "$@"
+  if (ipmitool chassis bootparam get 5 | grep -i 'CMOS Clear'>/dev/null); then
```
nest1ing, line 262: Probably `busctl` would also be better here.

commands/host.vegman
```diff
@@ -224,6 +224,90 @@ function cmd_virtualmedia_umount {
...
+  if [[ -z "${yes}" ]]; then
+    confirm
+  fi
+
+  ipmitool chassis bootdev none clear-cmos=yes >/dev/null
+}
+
+function config_default_cancel {
+  local yes=""
+  while [[ $# -gt 0 ]]; do
```
nest1ing: This code is excess too, by the same reason.

commands/host.vegman
AlexanderAmelkin: I'm fine with `ipmitool` here. It gives a more precise status because that's essentially what the BIOS will do.

commands/host.vegman
AlexanderAmelkin: `ipmitool` here also clears the other boot device settings, and I can't say that's bad for the 'reset to defaults' case.

nest1ing: Ok, let it be as is.

commands/host.vegman
nest1ing: Ok, as you believe that it's reasonable as let it be as is.

alatarum:
> maybe use `abort_badarg "$action"` here

It's not "bad arg", you normally can't get here because of value for action assigned in "case" above, arguments are not passed to action directly.
But we can use this helper just for simplicity.

commands/host.vegman
```diff
@@ -303,8 +289,8 @@ function cmd_config_default {
     cancel) config_default_cancel $yes;;
     status) config_default_status;;
     set)    config_default_set $yes;;
-    *) echo "Unsupported parameter. Use --status, --cancel or --yes to
-approve";;
+    *) echo "Unsupported parameter. Use --status, \
+             --cancel or --yes to approve";;
```
AlexanderAmelkin: Wrong. This will print a long string of spaces between "--status," and "--cancel".
You need to close the quote before `\` and then reopen it after indentaion. See my previous comment.

Also, why are there 3 commits now? I expected 2: one for the config_default, and one for help refactoring.

AlexanderAmelkin (changes requested):

commands/host.vegman
AlexanderAmelkin: Oh, and I totally missed that! Andrey was correct. You can never get to this line at all!
You've already directly set `action` in the above loop to one of the three values.
I'd remove this `case` completely and just call `config_default_${action} "${yes}"` here.

AlexanderAmelkin (changes requested):

AlexanderAmelkin (approved):

PR #36, opened 2021-06-04
bmc datetime set: Add range sanity checks

AlexanderAmelkin:
Only allow valid dates. The year range is limited 2021..2037.
The lower limit is obvious. The upper limit is because
mktime() in OpenBMC suffers from Y2k38 problem and won't accept
dates beyond 03:14:07 on Tuesday, 19 January 2038, so 2037
is the last year for which the full range of other components
is accepted.

End-user-impact: Only valid dates are now allowed

### DIFF
--- a/commands/bmc.nicole
+++ b/commands/bmc.nicole
@@ -152,14 +152,45 @@ function cmd_datetime_method {
 # Set system time manually
 #   YYYY-MM-DD-HH-mm-SS - Date and time to set
 function cmd_datetime_set {
+  local datestr="${1-}"
+
   # Only check format here, values are checked for sanity by `date`
-  if ! [[ ${1:-} =~ ^[0-9]{4}(-[0-9]{2}){5}$ ]]; then
+  if ! [[ "${datestr}" =~ ^[0-9]{4}(-[0-9]{2}){5}$ ]]; then
     echo "Please enter date and time in YYYY-MM-DD-HH-mm-SS format" >&2
     return 1
   fi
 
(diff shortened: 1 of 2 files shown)


commands/bmc.nicole
```diff
@@ -152,14 +152,45 @@ function cmd_datetime_method {
 # Set system time manually
 #   YYYY-MM-DD-HH-mm-SS - Date and time to set
 function cmd_datetime_set {
+  local datestr="${1-}"
```
nest1ing, line 155: Likely you meant `${1:-}`

commands/bmc.nicole
```diff
@@ -152,14 +152,45 @@ function cmd_datetime_method {
...
+  local YY
+  local MM
+  local DD
+  local HH
+  local mm
+  local SS
+  IFS='-' read YY MM DD HH mm SS < <(echo ${datestr})
+
+  # Problem Y2k38 isn't resolved in OpenBMC yet
+  if [[ $YY -le 2021 ]] || [[ $YY -gt 2037 ]]; then
```
nest1ing: Probably `-lt`. Or `2021` is also prohibited? ;)

AlexanderAmelkin: Fixed

nest1ing (approved):

commands/bmc.nicole
AlexanderAmelkin: They are the same, see https://infoheap.com/bash-set-default-value-if-a-variable-is-empty/

PR #40, opened 2021-06-16
host: Add 'nmi' command

nonameletters:
Added command 'nmi' to generate "Non-maskable interrupt" on host

End-user-impact: Added command to cli:
                 - `host nmi [-y | --yes]`
                   BMC generates NMI interrupt on host

Signed-off-by: Vladimir Mitrofanov <v.mitrofanov@yadro.com>

### DIFF
--- a/commands/host.vegman
+++ b/commands/host.vegman
@@ -289,6 +289,34 @@ function cmd_config_default {
   config_default_${action} ${yes}
 }
 
+# @sudo cmd_nmi admin
+# @restrict cmd_nmi admin
+# @doc cmd_nmi
+# Emit an NMI to the host
+#   -y, --yes - Do not ask for confirmation
+function cmd_nmi {
+  local yes=""
+
+  while [[ $# -gt 0 ]]; do
+    case "${1}" in
+      -y | --yes) yes="y";;
+      *) abort_badarg "${1}";;
+    esac
+    shift
+  done
+
+  echo "An NMI signal will be now sent to the host."
+  echo "That may cause host OS crash an reboot if configured so."
+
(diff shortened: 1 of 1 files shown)


commands/host.vegman
```diff
@@ -289,6 +289,31 @@ function cmd_config_default {
   config_default_${action} ${yes}
 }
 
+# @sudo cmd_nmi admin
+# @restrict cmd_nmi admin
+# @doc cmd_nmi
+# Emmit NMI interrupt
```
AlexanderAmelkin: `Emit`. Single m. Also an 'an' article is missing, and I in NMI already stands for 'Interrupt', so:
`Emit an NMI to the host`

AlexanderAmelkin (changes requested):

commands/host.vegman
```diff
@@ -289,6 +289,31 @@ function cmd_config_default {
...
+
+  while [[ $# -gt 0 ]]; do
+    case "${1}" in
+      -y | --yes) yes="y";;
+      *) abort_badarg "${1}";;
+    esac
+    shift
+  done
+
+  if [[ -z "${yes}" ]]; then
```
AlexanderAmelkin, line 311: It would be nice to print something like:
```
An NMI signal will now be sent to the host.
That may cause host OS crash an reboot if configured so.
```

I would print that before the `if` so that it would be displayed always. If `--yes` is not specified, then the "Are you sure?" question will appear right after that message, which looks like a nice place for it to me.

AlexanderAmelkin (changes requested):

AlexanderAmelkin (approved):

PR #42, opened 2021-06-23
host: Update 'virtualmedia ' command

ikmsk10:
By the new VirtualMedia implementation, the CLI commands have been
redesigned.

End-user-impact: Update the following commands of CLI:

- ` host virtualmedia mount`
   Virtual Media possible to mount image over SMB and  HTTPS
```bash
  --usb      - The USB interface type
  --usb-ro   - The USB interface type (read-only)
  --hdd      - The HDD interface type
  --cdrom    - The CD/DVD/BD-ROM interface type (read-only), based on file size
  FILE       - Path to the image file. Supported the SMB and the HTTPS protocols.
 -r          - Readonly mode
 -u username - Username credential to connect SMB or HTTPS remote server.
 -p          - Request to input the password credential to connect SMB or HTTPS remote server.
```   

- `host virtualmedia umount`
VirtualMedia able to unmount WebUI-Proxy, CLI-mounted devices
```bash
  MEDIA     - the Virtual Media Device name to unmount (USB0 or USB1)
  -y, --yes - Do not ask for confirmation
```  
    
Add following commands to CLI:
- `host virtualmedia status`
   Get status of the Virtual Media devices
```
Virtual Media (USB0) status:	mounted
	Image: ru_windows7_SP1_x64_Enterprise_2020_dvd.iso
Virtual Media (USB1) status:	mounted
	Image: https://ftp.uni-erlangen.de/mirrors/ubuntu-releases/20.04/ubuntu-20.04.2.0-desktop-amd64.iso
```
---
	
Remove the `bmc remoteimage` command as no longer supported in the CLI.

```

### CODE REVIEW

File: commands/host.vegman [hunk 1]
```diff
@@ -158,25 +158,70 @@ function cmd_virtualmedia {
...
+
+  for usb_item_metadata in "${usb_status_metadata[@]}"
+  do
+    local id="${usb_item_metadata%%;*}"
+    local vm_type="${usb_item_metadata#*;}"
+    local usb_status=$(busctl get-property xyz.openbmc_project.VirtualMedia \
+                                        /xyz/openbmc_project/VirtualMedia/${vm_type}/${id} \
+                                        xyz.openbmc_project.VirtualMedia.Process Active)
+    local image=""
+    echo -n "The (USB0) Virtual Media status: "
```
nest1ing: - `The` seems to me excess here.
 - `USB0` looks like a missed hadrcode
 - If the tab character is really required as a delimiter put it in this statement.

So, I propose to change this line to something like:
```bash
echo -ne "Virtual Media (${id}) status:\t"
```

commands/host.vegman
```diff
@@ -158,25 +158,70 @@ function cmd_virtualmedia {
...
+    if [ "$usb_status" == "b true" ]; then
+      echo -e "\tmounted"
```
nest1ing: And then here would be just `echo "mounted"`

commands/host.vegman
```diff
@@ -193,35 +238,64 @@ function cmd_virtualmedia_mount {
...
 
   if [[ -z "${yes}" ]]; then
     confirm
   fi
 
-  usb-ctrl eject "usb.ms.cli"
+  busctl call xyz.openbmc_project.VirtualMedia \
+    "/xyz/openbmc_project/VirtualMedia/${dbus_vm_type}/${media}" \
+    "xyz.openbmc_project.VirtualMedia.${dbus_vm_type}" Unmount &>2 && \
+  echo "The ${media} Virtual Media Device unmounted"
```
nest1ing: Here also `The` looks excess.
`Virtual Media Device ${media} unmounted`

commands/host.vegman
```diff
@@ -193,35 +238,64 @@ function cmd_virtualmedia_mount {
...
+  then
+    printf %b "${username:0:64}\0${password:0:64}\0" >6
+    cred_pipe=6
+  fi
+
+  $(busctl call xyz.openbmc_project.VirtualMedia \
+    /xyz/openbmc_project/VirtualMedia/Legacy/USB1 \
+    xyz.openbmc_project.VirtualMedia.Legacy \
+    Mount sbyv "$imagefile" $readonly ${type_id} y $cred_pipe  >/dev/null) && \
+  echo "The USB1 Virtual Media Device mounted"
```
nest1ing: And here: `Virtual Media Device USB1 mounted`.
And what about USB0? How can it be mounted?

commands/host.vegman
ikmsk10: Yes, `USB0` should become `${id}`.

commands/host.vegman
ikmsk10: Now. the system has the `virtual-media` service, which defines each `VM-endpoint` as `DBus` object.
The VEGMAN system has a tow VM-endpoints: WebUI Proxy (USB0) and Legacy(USB1).
The USB0 can be mounted only via WebUI because it uses the NBD-proxy-server. In the CLI we can only get the status of the `USB0` and unmount it.
For the USB1 we possible to `mount`/`umount`/`status` via `CLI`

commands/host.vegman
```diff
@@ -193,35 +238,64 @@ function cmd_virtualmedia_mount {
...
-    echo "Please specify the path to image file" >&2
+  if ! [[ ${imagefile} =~ ${imagefile_pattern} ]]; then
+    echo "Please specify the valid path to the image file" >&2
     return 1
   fi
 
-  usb-ctrl insert "usb.ms.cli" "${imagefile}" "${type}"
+  if [[ ! -z ${username} ]] || [[ ! -z ${password} ]]
+  then
+    printf %b "${username:0:64}\0${password:0:64}\0" >6
```
ikmsk10: @nest1ing, I very worry about packing and secure passing credentials via PIPE.
The virtual media expect blob `%{username}\0%{password}`, but is there possible a case when `printf  %b` might cut blob by first `\0`?

nest1ing: `%b` is a bash-specific modifier. I assume it will work as you expect and will not truncate the result string.
But if you doubt replace it with the standard C-style printf format:
```bash
printf "%s\0%s\0" "${username:0:64}" "${password:0:64}" >6
```

nest1ing: There is a thing makes me confused a little bit more than `printf` format string:
What is `6`? It looks like a file name, but not a PIPE.
Who should receive this data?

AlexanderAmelkin: Yes, there is such a case. If `${password}` starts with a number, it will get appended to `\0` and will not be handled correctly:

```
$ pwd=456
$ printf %b "123\0${pwd}" | hexdump -C                                                                                        
00000000  31 32 33 2e                                       |123.|
00000004
$ pwd=abc
$ printf %b "123\0${pwd}" | hexdump -C                                                                                        
00000000  31 32 33 00 61 62 63                              |123.abc|
00000007
```

ikmsk10: Yes, it looks like a mistake.
The `FD` of PIPE should be passed to the DBus `Mount` method, and the `virtual media` service will accept the PIPE and read credentials via it.
I guess I should open a new pipe first as FIFO, associate it to some tmpfile, rm file, and after correct read, do unlink FD.

ikmsk10: fixed

commands/host.vegman
ikmsk10: fixed

commands/host.vegman
ikmsk10: fixed

commands/host.vegman
ikmsk10: fixed

commands/host.vegman
```diff
@@ -158,25 +158,66 @@ function cmd_virtualmedia {
...
+    echo -ne "Virtual Media ($id) status:\t"
+    if [ "$usb_status" == "b true" ]; then
+      echo "mounted"
+      image=$(busctl get-property  xyz.openbmc_project.VirtualMedia \
+                                  /xyz/openbmc_project/VirtualMedia/${vm_type}/${id} \
+                                  xyz.openbmc_project.VirtualMedia.MountPoint ImageURL)
+      image=${image#*\"}
+      echo -e "\tImage: ${image%\"*}"
+    else
+      echo -e "unmounted"
```
nest1ing: `-e` is no longer required here.

commands/bmc.vegman
```diff
@@ -406,114 +406,3 @@ function cmd_reboot {
   echo "Reboot the BMC..."
   reboot ${force}
 }
-
```
AlexanderAmelkin, line 409: I believe this must go in a separate commit saying `bmc: Remove 'remoteimage' command`

commands/host.vegman
```diff
@@ -158,25 +158,66 @@ function cmd_virtualmedia {
   subcommand "$@"
 }
 
+# @sudo cmd_virtualmedia_status admin
+# @restrict cmd_virtualmedia_status admin
+# @doc cmd_virtualmedia_status
+# Get status of the virtual media
```
AlexanderAmelkin: `Get the status of virtual media`. Media is plural, it can go without an article unless you're talking about some very specific media, status is singular and so it needs an article.

commands/host.vegman
```diff
@@ -158,25 +158,66 @@ function cmd_virtualmedia {
...
+function cmd_virtualmedia_status {
+  local usb_status_metadata=( "USB0;Proxy" "USB1;Legacy" )
```
AlexanderAmelkin: this must be `local -a` if you want a bash array

commands/host.vegman
alatarum: I'd like you to mention this somewhere in help message.

alatarum (dismissed):

commands/host.vegman
```diff
@@ -158,25 +158,66 @@ function cmd_virtualmedia {
...
+
+  for usb_item_metadata in "${usb_status_metadata[@]}"
+  do
+    local id="${usb_item_metadata%%;*}"
+    local vm_type="${usb_item_metadata#*;}"
```
AlexanderAmelkin: why `id` was greedy (`%%`), but this one isn't (`#` instead of `##`) ?
Doesn't make any actual difference in any of the cases, but still.

commands/host.vegman
```diff
@@ -158,25 +158,66 @@ function cmd_virtualmedia {
...
+                                        xyz.openbmc_project.VirtualMedia.Process Active)
+    local image=""
+    echo -ne "Virtual Media ($id) status:\t"
+    if [ "$usb_status" == "b true" ]; then
+      echo "mounted"
+      image=$(busctl get-property  xyz.openbmc_project.VirtualMedia \
+                                  /xyz/openbmc_project/VirtualMedia/${vm_type}/${id} \
+                                  xyz.openbmc_project.VirtualMedia.MountPoint ImageURL)
+      image=${image#*\"}
+      echo -e "\tImage: ${image%\"*}"
```
AlexanderAmelkin: This outputs:
```
Virtual Media (USB0) status: mounted
        Image: protocol://some.domain.tld/path/to/image.bin
```

I don't see why this `\t` is there. It just consumes space. I guess this reads better:
```
Virtual Media (USB0) status: mounted
Image: protocol://some.domain.tld/path/to/image.bin
```
Please remove `\t`.

commands/bmc.vegman
AlexanderAmelkin: Also, I don't see any way to mount a remote NBD with the updated `host virtualmedia` command.
It was used to mount remote devices, not just files. I don't think we need to remove this command if we don't provide a real alternative to it.

AlexanderAmelkin (changes requested):

commands/host.vegman
```diff
@@ -193,35 +234,75 @@ function cmd_virtualmedia_mount {
...
-  usb-ctrl insert "usb.ms.cli" "${imagefile}" "${type}"
+  if [[ ! -z ${username} ]] || [[ ! -z ${password} ]]
+  then
+    local cred_pipe_fifo=$(mktemp -u)
+    cred_pipe=6
+    mkfifo $cred_pipe_fifo
+    exec 6<>$cred_pipe_fifo
+    rm $cred_pipe_fifo
+
+    printf %b "${username:0:64}\0${password:0:64}\0\n" >&6
```
nest1ing: As @AlexanderAmelkin  noticed the piped data will be corrupted if `password` begins with a digit.
And as it's shown by my tests the C-style printf format works properly even with a numeric `password`.

commands/host.vegman
```diff
@@ -193,35 +234,75 @@ function cmd_virtualmedia_mount {
...
     return 1
   fi
 
-  usb-ctrl insert "usb.ms.cli" "${imagefile}" "${type}"
+  if [[ ! -z ${username} ]] || [[ ! -z ${password} ]]
+  then
+    local cred_pipe_fifo=$(mktemp -u)
+    cred_pipe=6
+    mkfifo $cred_pipe_fifo
+    exec 6<>$cred_pipe_fifo
```
nest1ing: You don't need to hardcode this value. The line 245 is excess.
Use:
```bash
exec {cred_pipe}<>${cred_pipe_fifo}
```
and `${cred_pipe}` will contain the actual value.

commands/host.vegman
```diff
@@ -193,35 +234,75 @@ function cmd_virtualmedia_mount {
...
+
+  $(busctl call xyz.openbmc_project.VirtualMedia \
+    /xyz/openbmc_project/VirtualMedia/Legacy/USB1 \
+    xyz.openbmc_project.VirtualMedia.Legacy \
+    Mount sbyv "$imagefile" $readonly ${type} y $cred_pipe  >/dev/null) && \
+  echo "Virtual Media (USB1) mounted" || \
+  echo "Fail to mount Virtual Media (USB1)"
+
+  if [[ $cred_pipe -eq 6 ]]; then
+    exec 6>&-
```
nest1ing: The same as during opening:
```bash
exec {cred_pipe}>&-
```

nest1ing (dismissed):

commands/host.vegman
```diff
@@ -193,35 +234,75 @@ function cmd_virtualmedia_mount {
...
+    exec 6<>$cred_pipe_fifo
+    rm $cred_pipe_fifo
+
+    printf %b "${username:0:64}\0${password:0:64}\0\n" >&6
+  fi
+
+  $(busctl call xyz.openbmc_project.VirtualMedia \
+    /xyz/openbmc_project/VirtualMedia/Legacy/USB1 \
+    xyz.openbmc_project.VirtualMedia.Legacy \
+    Mount sbyv "$imagefile" $readonly ${type} y $cred_pipe  >/dev/null) && \
```
nest1ing: But I still don't understand how can it work?
`virtual-media` is not our child process and it knows nothing about our local file descriptors.
Probably, we should specify here the full path to the named pipe.

nest1ing: OK, I found out that it is an internal `systemd` possibilities. There is a special D-Bus type `unix_fd` that can be passed between client and service. And `libsystemd` likely handles it correct.

Let's assume that it works this way.

commands/bmc.vegman
ikmsk10: After discussing the TG, we decided to implement mounting the system-based device as the Virtual Media image in the next revision. Is I understand right?

commands/host.vegman
ikmsk10: Yes, in this case, it doesn't matter. I guess this is just a mistake. To uniform, I'll add `##` to grep `vm_type`.

commands/host.vegman
ikmsk10: yes, I know, but 
```
:~$ pipe_id=6
:~$ cred_pipe_fifo=$(mktemp -u)
:~$ mkfifo $cred_pipe_fifo
:~$ exec $pipe_id<>$cred_pipe_fifo
bash: exec: 6: not found
```
and `exec 6<>$cred_pipe_fifo` is success

ikmsk10: Ahh, now I have seen. Should be `{cred_pipe}`, not `${cred_pipe}`

ikmsk10: I checked script with `exec {cred_pipe}<>${cred_pipe_fifo}` and the suggested way doesn't works.

commands/bmc.vegman
AlexanderAmelkin: Yes

commands/host.vegman
AlexanderAmelkin: I just checked. I created a file `x`:

```
#!/bin/bash
cred_pipe_fifo=$(mktemp -u)
mkfifo $cred_pipe_fifo
exec {cred_pipe}<>${cred_pipe_fifo}
echo Fifo is at ${cred_pipe}
```

and ran it:

```
$ chmod 755 ./x
$ ./x
Fifo is at 10
$
```

commands/host.vegman
AlexanderAmelkin: I confirm that this works as expected:

```
$ printf "%s\0%s\0\n" username password | hexdump -C                                                                            
00000000  75 73 65 72 6e 61 6d 65  00 70 61 73 73 77 6f 72  |username.passwor|
00000010  64 00 0a                                          |d..|
```

commands/host.vegman
ikmsk10: I have a little bit confused... For a bash terminal, it's doesn't worked. But in the script is it.
Ok, I changed `6` to `{cred_pipe}`.

commands/host.vegman
ikmsk10: Done, now the output is
```
vegman-s320-2007210001:~$ host virtualmedia status           
Virtual Media (USB0) status: unmounted
Virtual Media (USB1) status: mounted
Image: https://ftp.uni-erlangen.de/mirrors/ubuntu-releases/20.04/ubuntu-20.04.2.0-desktop-amd64.iso

```

commands/host.vegman
ikmsk10: done

commands/host.vegman
ikmsk10: Fixed

commands/host.vegman
ikmsk10: Yes, changed as same above

commands/host.vegman
ikmsk10: fixed

commands/host.vegman
ikmsk10: Fixed

commands/host.vegman
ikmsk10: Changed

AlexanderAmelkin (approved):

PR #43, opened 2021-06-29
Remove manual datetime settings

nest1ing:


commands/bmc.vegman
```diff
@@ -106,7 +106,7 @@ function cmd_datetime {
 # @restrict cmd_datetime_method admin
 # @doc cmd_datetime_method
 # Change time synchronization method
-#   {manual|ntp} - The synchronization method to use
+#   {HostTime|NTP} - The synchronization method to use
```
AlexanderAmelkin: Please avoid CamelCase in CLI commands and their arguments.
Also, let's make it `host|ntp`.

commands/bmc.vegman
```diff
@@ -116,12 +116,17 @@ function cmd_datetime_method {
...
+  fi
+
   if [[ $# -eq 0 ]]; then
       echo "Current time synchronization method is ${method}"
   else
     case "${1}" in
-      [Mm][Aa][Nn][Uu][Aa][Ll])
-        if [[ ${method} != "Manual" ]]; then
+      [Hh][Oo][Ss][Tt][Tt][Ii][Mm][Ee])
+        if [[ ${method} != "HostTime" ]]; then
```
alatarum: That's look strange: why do you need to translate dbus property value from "Manual" to "HostTime" and then back. I get it as you want to change CLI command argument, but this doesn't require to also change property value.
From first look this "if [[ ${method} != "HostTime" ]]; then ... set ... Method.Manual" looks as error and it took me some time to understand what's going on here.

commands/bmc.vegman
```diff
@@ -116,12 +116,17 @@ function cmd_datetime_method {
...
+      method="HostTime"
+  fi
+
   if [[ $# -eq 0 ]]; then
       echo "Current time synchronization method is ${method}"
   else
     case "${1}" in
-      [Mm][Aa][Nn][Uu][Aa][Ll])
-        if [[ ${method} != "Manual" ]]; then
+      [Hh][Oo][Ss][Tt][Tt][Ii][Mm][Ee])
```
AlexanderAmelkin: better use `${1,,}` and compare to a lowercase string.

AlexanderAmelkin (changes requested):

commands/bmc.vegman
nest1ing: API changing affects to many other software (settings, phosphor-time-manager, bmcweb, webui and so on). So, this masquerade seems to me a little bit easier to implement.

commands/bmc.vegman
nest1ing: done

commands/bmc.vegman
nest1ing: done

commands/bmc.vegman
nest1ing: Oh, and the property value is displayed to user on line 125.

alatarum: ok

alatarum: But you can have this masquerading 'if' between lines 124 and 125

nest1ing: I have got what do you meant... likely... 
fixed :)

alatarum (dismissed):

commands/bmc.vegman
```diff
@@ -117,10 +117,11 @@ function cmd_datetime_method {
   method=${method%\"*}
 
   if [[ $# -eq 0 ]]; then
-      echo "Current time synchronization method is ${method}"
+      # VEGMAN doesn't have RTC and in manual mode it uses PCH time.
+      echo "Current time synchronization method is ${method/Manual/HostTime}"
```
AlexanderAmelkin: So now we have it like this?

```
$ bmc datetime method host
$ bmc datetime method
Current time synchronization method is HostTime
```

Doesn't this looks strange to you?
Let's make it `Host` here so that it matches the command argument.

AlexanderAmelkin (changes requested):

commands/bmc.vegman
nest1ing: done

AlexanderAmelkin (approved):

alatarum (approved):

PR #44, opened 2021-07-05
host: virtual-media: Support the HTTP protocol.

ikmsk10:
This patch brings support of HTTP protocol for mount VM-image from
external server.

End-user-impact: Now, the `host virtualmedia mount` command supports the
                 HTTP protocol to mount Virtual Media images from an
                 external server via HTTP.

### DIFF
--- a/commands/host.vegman
+++ b/commands/host.vegman
@@ -192,14 +192,14 @@ function cmd_virtualmedia_status {
 # @restrict cmd_virtualmedia_mount admin
 # @doc cmd_virtualmedia_mount
 # Mount a virtual media
-#   --usb      - The USB interface type
-#   --usb-ro   - The USB interface type (read-only)
-#   --hdd      - The HDD interface type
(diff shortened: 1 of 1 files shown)


commands/host.vegman
```diff
@@ -196,17 +196,17 @@ function cmd_virtualmedia_status {
...
+#  -u username - Username credential to connect SMB or HTTP(S) remote server.
+#  -p          - Request to input the password credential to connect SMB or HTTP(S) remote server.
 function cmd_virtualmedia_mount {
   local imagefile=""
   local type=0
   local readonly=false
   local username=""
   local password=""
-  local imagefile_pattern="^(https|smb)"
+  local imagefile_pattern="^(http:|https:|smb:)"
```
nest1ing: It doesn't look wrong but I would truncate it to something like: `"^(https?|smb):"`

nest1ing (dismissed):

PR #55, opened 2021-09-13
vegman: intrusion: add intrusion control

fr0st61te:
Add intrusion control interface which described in CLI documentation.

Signed-off-by: Ivan Mikhaylov <i.mikhaylov@yadro.com>

### DIFF
--- a/commands/diagnostics.vegman
+++ b/commands/diagnostics.vegman
@@ -119,3 +119,87 @@ function cmd_traceroute {
 function cmd_netstat {
   netstat "$@"
 }
+
+# D-Bus intrusion objects
+INTR_SERVICE="xyz.openbmc_project.intrusionsensor.service"
+INTR_RST_GPIO_PIN="CASE_OPEN_FAULT_RST"
+INTR_DBUS_SERVICE="xyz.openbmc_project.IntrusionSensor"
+INTR_DBUS_PATH="/xyz/openbmc_project/Intrusion/Chassis_Intrusion"
+INTR_DBUS_INTERFACE="xyz.openbmc_project.Chassis.Intrusion"
+INTR_DBUS_PROPERTY="Status"
+
+# @sudo cmd_intrusion_on admin,tech
+# @restrict cmd_intrusion_on admin,tech
+# @doc cmd_intrusion_on
+# Turn the chassis intrusion detection on
+function cmd_intrusion_on {
+  systemctl start ${INTR_SERVICE}
+  systemctl enable ${INTR_SERVICE}
+}
+
+# @sudo cmd_intrusion_off admin,tech
+# @restrict cmd_intrusion_off admin,tech
+# @doc cmd_intrusion_off
+# Turn the chassis intrusion detection off
+function cmd_intrusion_off {
+  systemctl stop ${INTR_SERVICE}
+  systemctl disable ${INTR_SERVICE}
+}
+
+get_intrusion_state() {
+  busctl get-property ${INTR_DBUS_SERVICE} "${INTR_DBUS_PATH}" \
+                      ${INTR_DBUS_INTERFACE} ${INTR_DBUS_PROPERTY} \
+                      2>/dev/null | awk '{print $NF;}' ||:
+}
+
+# @sudo cmd_intrusion_show admin,tech
+# @restrict cmd_intrusion_show admin,tech
+# @doc cmd_intrusion_show
+# Show the chassis intrusion detection status
+function cmd_intrusion_show {
+  state=$(get_intrusion_state)
+  if [[ -z "$state" ]]; then
+    echo "Intrusion sensor service is stopped or failed"
+  elif [ "$state" == "Normal" ]; then
+    echo "Chassis is closed"
+  else
+    echo "Hardware intrusion detected"
+  fi
+}
+
+# @sudo cmd_intrusion_clear admin,tech
+# @restrict cmd_intrusion_clear admin,tech
+# @doc cmd_intrusion_clear
+# Clear the chassis intrusion status
+function cmd_intrusion_clear {
+  local yes=""
+  while [[ $# -gt 0 ]]; do
+    case "$1" in
+      -y | --yes) yes="y";;
+      *) abort_badarg "$1";;
+    esac
+    shift
+  done
+
+  pin=$(gpiofind ${INTR_RST_GPIO_PIN})
+  if [[ -z "${pin}" ]]; then
+    echo "Intrusion detection reset is not available for this product"
+    return 1
+  fi
+
+  if [[ -z "${yes}" ]]; then
+    confirm
+  fi
+
+  state=$(get_intrusion_state)
+  if [[ -z "$state" ]]; then
+    state="n/a"
+  fi
+
+  #shellcheck disable=SC2086
+  gpioset -m wait ${pin}=1 && \
+  log_audit "Asserted chassis intrusion reset pin from state: $state"
+
+  #shellcheck disable=SC2086
+  gpioset ${pin}=0
+}
--- a/commands/diagnostics.vegman
+++ b/commands/diagnostics.vegman
@@ -27,7 +27,7 @@ get_led_state() {
     local name=$1
     busctl get-property ${LED_DBUS_SERVICE} "${LED_DBUS_PATH}/${name}" \
                         ${LED_DBUS_INTERFACE} ${LED_DBUS_PROPERTY} \
-                        | awk '{print $NF}' 2>/dev/null
+                        2>/dev/null | awk '{print $NF}' ||:
 }
 
 set_led_state() {
@@ -36,7 +36,7 @@ set_led_state() {
 
     busctl set-property ${LED_DBUS_SERVICE} "${LED_DBUS_PATH}/${name}" \
                         ${LED_DBUS_INTERFACE} ${LED_DBUS_PROPERTY} \
-                        b "${state}" 2>/dev/null
+                        b "${state}" 2>/dev/null ||:
 }
 
 # @sudo cmd_led_status admin,operator,user
@@ -51,12 +51,12 @@ function cmd_led_status {
   expect_noarg "$@"
   local state
 
-  state=$(get_led_state ${ID_LED_BLINKING} || :)
+  state=$(get_led_state ${ID_LED_BLINKING})
 
   # `enclosure_identify` has no sense while `enclosure_identify_blink` is
   # asserted.
   if [[ "${state}" != "true" ]]; then
-    state=$(get_led_state ${ID_LED_ON} || :)
+    state=$(get_led_state ${ID_LED_ON})
   fi
 
   case "${state}" in
@@ -74,7 +74,7 @@ function cmd_led_status {
 function cmd_led_on {
   expect_noarg "$@"
 
-  set_led_state ${ID_LED_BLINKING} false ||:
+  set_led_state ${ID_LED_BLINKING} false
   set_led_state ${ID_LED_ON} true
 
   # cmd_led_status takes no arguments, silience shellcheck about it
@@ -88,7 +88,7 @@ function cmd_led_on {
 # Turn the identification LED off
 function cmd_led_off {
   expect_noarg "$@"
-  set_led_state ${ID_LED_BLINKING} false ||:
+  set_led_state ${ID_LED_BLINKING} false
   set_led_state ${ID_LED_ON} false
 
   # cmd_led_status takes no arguments, silience shellcheck about it


commands/diagnostics.vegman
```diff
@@ -119,3 +119,50 @@ function cmd_traceroute {
...
+  fi
+}
+
+# @sudo cmd_intrusion_clear admin,tech
+# @restrict cmd_intrusion_clear admin,tech
+# @doc cmd_intrusion_clear
+# Clear the chassis intrusion status
+function cmd_intrusion_clear {
+  # GPIO pin A1 = 1
+  gpioset gpiochip0 1=1 && log_audit "Asserted chassis intrusion pin"
```
alatarum: Please, use gpiofind here.
Also, keep in mind, this will only work for N110/Sx20, but not for Rx20.
BTW, isn't RST pin should be de-asserted to initial state?

commands/diagnostics.vegman
```diff
@@ -119,3 +119,50 @@ function cmd_traceroute {
...
+INTR_DBUS_PATH="/xyz/openbmc_project/Intrusion/Chassis_Intrusion"
+INTR_DBUS_INTERFACE="xyz.openbmc_project.Chassis.Intrusion"
+INTR_DBUS_PROPERTY="Status"
+
+# @sudo cmd_intrusion_on admin,tech
+# @restrict cmd_intrusion_on admin,tech
+# @doc cmd_intrusion_on
+# Turn the chassis intrusion detection on
+function cmd_intrusion_on {
+  sysctemctl start ${INTR_SERVICE}
```
alatarum: Shouldn't you also make enable/disable to make this reboot-persistent?
Also, I think we should call cmd_intrusion_clear here before service start (IMHO to get outdated intrusion warning right after manual sensor activation is unexpected behavior for end user).

commands/diagnostics.vegman
```diff
@@ -119,3 +119,50 @@ function cmd_traceroute {
...
+# @doc cmd_intrusion_show
+# Show the chassis intrusion detection status
+function cmd_intrusion_show {
+  state=$(busctl get-property ${INTR_DBUS_SERVICE} "${INTRUSION_DBUS_PATH}" \
+          ${INTR_DBUS_INTERFACE} ${INTRUSION_DBUS_PROPERTY} \
+          | awk '{print $NF;}')
+  if [ "$state" == "Normal" ]; then
+    echo "Chassis is closed"
+  else
+    echo "Hardware intrusion"
```
alatarum: "Hardware intrusion *detected*"?

commands/diagnostics.vegman
```diff
@@ -119,3 +119,50 @@ function cmd_traceroute {
...
+# @sudo cmd_intrusion_show admin,tech
+# @restrict cmd_intrusion_show admin,tech
+# @doc cmd_intrusion_show
+# Show the chassis intrusion detection status
+function cmd_intrusion_show {
+  state=$(busctl get-property ${INTR_DBUS_SERVICE} "${INTRUSION_DBUS_PATH}" \
+          ${INTR_DBUS_INTERFACE} ${INTRUSION_DBUS_PROPERTY} \
+          | awk '{print $NF;}')
+  if [ "$state" == "Normal" ]; then
+    echo "Chassis is closed"
```
alatarum, line 164: For Rx20 this statement correct, but for N110/Sx20 it more like "Hardware intrusion was not detected", since chassis can be already closed again, but intrusion remembered.

commands/diagnostics.vegman
fr0st61te: Any idea how to distinguish between boards, only with proper GPIO pin naming?

commands/diagnostics.vegman
fr0st61te: I think user should clear it, not automatically by any means.

commands/diagnostics.vegman
nest1ing: I agree with @alatarum. In additional I suggest to specify the impulse duration (`-m wait -u 500000`)

fr0st61te: > isn't RST pin should be de-asserted to initial state?

no, as I see from doc.

commands/diagnostics.vegman
nest1ing: I guess we should also handle the case when the tool fails.

commands/diagnostics.vegman
alatarum: > Any idea how to distinguish between boards, only with proper GPIO pin naming?

No, have no idea... But you can use /etc/os-release

> no, as I see from doc.

What doc? What I see in task description, it says "assert for 100-500ms". I believe, it should be cleared, otherwise, how  you gonna to asset it next time?

commands/diagnostics.vegman
alatarum: @AlexanderAmelkin , what do you think?

commands/diagnostics.vegman
alatarum: > No, have no idea... But you can use /etc/os-release

But, well... Maybe just use gpiofind to locate RST pin and if nothing found - then say that it is not supported.

fr0st61te: Yea, gpiofind may distinguish it, rely on /etc/os-release seems not so good...

commands/diagnostics.vegman
```diff
@@ -119,3 +119,57 @@ function cmd_traceroute {
...
+}
+
+# @sudo cmd_intrusion_show admin,tech
+# @restrict cmd_intrusion_show admin,tech
+# @doc cmd_intrusion_show
+# Show the chassis intrusion detection status
+function cmd_intrusion_show {
+  state=$(busctl get-property ${INTR_DBUS_SERVICE} "${INTR_DBUS_PATH}" \
+          ${INTR_DBUS_INTERFACE} ${INTR_DBUS_PROPERTY} \
+          | awk '{print $NF;}' 2>/dev/null ||:)
```
nest1ing: stderr redirection should be applied for busctl call not for awk. ;)

commands/diagnostics.vegman
```diff
@@ -119,3 +119,57 @@ function cmd_traceroute {
...
+  fi
+}
+
+# @sudo cmd_intrusion_clear admin,tech
+# @restrict cmd_intrusion_clear admin,tech
+# @doc cmd_intrusion_clear
+# Clear the chassis intrusion status
+function cmd_intrusion_clear {
+  # GPIO pin A1 = 1
+  pin=$(gpiofind ${INTR_GPIO_PIN} | awk '{print $NF;}')
```
nest1ing: `awk` is excess here. `gpiofind` returns the data used in the `gpioget`/`gpioset` command without any modifications. But, probably, it have sense to suppress `stderr` and prevent `bash -e` terminating on failures, like with `busctl` above.

commands/diagnostics.vegman
```diff
@@ -119,3 +119,57 @@ function cmd_traceroute {
...
+  if [[ -n "${pin}" ]]; then
+    gpioset gpiochip0 -m wait -u 250000 "${pin}"=1 && \
+    log_audit "Asserted chassis intrusion pin"
+
+    gpioset gpiochip0 "${pin}"=0
```
nest1ing: You were right, `gpioset -m wait` must restore the original GPIO state after exiting.
But I don't see any wrong in forcing pull it down. Let it be as is.

commands/diagnostics.vegman
nest1ing: I agree with @alatarum, on/off should be a combination of `systemctl enable/disable ...` and `systemctl start/stop ...` commands.
About auto resetting, I guess it shouldn't to do.

commands/diagnostics.vegman
```diff
@@ -119,3 +119,57 @@ function cmd_traceroute {
...
+
+# @sudo cmd_intrusion_clear admin,tech
+# @restrict cmd_intrusion_clear admin,tech
+# @doc cmd_intrusion_clear
+# Clear the chassis intrusion status
+function cmd_intrusion_clear {
+  # GPIO pin A1 = 1
+  pin=$(gpiofind ${INTR_GPIO_PIN} | awk '{print $NF;}')
+  if [[ -n "${pin}" ]]; then
+    gpioset gpiochip0 -m wait -u 250000 "${pin}"=1 && \
```
nest1ing: And as this operation has high security matter, it probably has a sense to prepend this with user confirmation and an optional key `-y/--yes` for using in the batch mode.

commands/diagnostics.vegman
```diff
@@ -119,3 +119,57 @@ function cmd_traceroute {
 function cmd_netstat {
   netstat "$@"
 }
+
+# D-Bus intrusion objects
+INTR_SERVICE="xyz.openbmc_project.intrusionsensor.service"
+INTR_GPIO_PIN="CASE_OPEN_FAULT_RST"
```
alatarum: I would name it INTR_RST_GPIO_PIN

commands/diagnostics.vegman
```diff
@@ -119,3 +119,57 @@ function cmd_traceroute {
...
+# @sudo cmd_intrusion_clear admin,tech
+# @restrict cmd_intrusion_clear admin,tech
+# @doc cmd_intrusion_clear
+# Clear the chassis intrusion status
+function cmd_intrusion_clear {
+  # GPIO pin A1 = 1
+  pin=$(gpiofind ${INTR_GPIO_PIN} | awk '{print $NF;}')
+  if [[ -n "${pin}" ]]; then
+    gpioset gpiochip0 -m wait -u 250000 "${pin}"=1 && \
+    log_audit "Asserted chassis intrusion pin"
```
alatarum: "intrusion RESET pin"
In Redfish spec this called `IntrusionSensorReArm`

commands/diagnostics.vegman
fr0st61te: Oops :D

commands/diagnostics.vegman
fr0st61te: Ok, I get it.

commands/diagnostics.vegman
```diff
@@ -119,3 +119,76 @@ function cmd_traceroute {
...
+# Turn the chassis intrusion detection off
+function cmd_intrusion_off {
+  systemctl stop ${INTR_SERVICE}
+  systemctl disable ${INTR_SERVICE}
+}
+
+get_intrusion_state() {
+  busctl get-property ${INTR_DBUS_SERVICE} "${INTR_DBUS_PATH}" \
+                      ${INTR_DBUS_INTERFACE} ${INTR_DBUS_PROPERTY} \
+                      | awk '{print $NF;}' 2>/dev/null
```
nest1ing: "Oops" is still here ;)

commands/diagnostics.vegman
```diff
@@ -119,3 +119,76 @@ function cmd_traceroute {
...
+}
+
+# @sudo cmd_intrusion_show admin,tech
+# @restrict cmd_intrusion_show admin,tech
+# @doc cmd_intrusion_show
+# Show the chassis intrusion detection status
+function cmd_intrusion_show {
+ state=$(get_intrusion_state || :)
```
nest1ing: I believe `||:` should be also moved into `get_intrusion_state`

commands/diagnostics.vegman
nest1ing: `[ -z "${state}" ]` should also be handled as the case when the service stopped/failed.

commands/diagnostics.vegman
```diff
@@ -119,3 +119,76 @@ function cmd_traceroute {
...
+  local yes=""
+  while [[ $# -gt 0 ]]; do
+    case "$1" in
+      -y | --yes) yes="y";;
+      *) abort_badarg "$1";;
+    esac
+    shift
+  done
+
+  if [[ -z "${yes}" ]]; then
```
nest1ing, line 190: The confirmation request is missing.

commands/diagnostics.vegman
```diff
@@ -119,3 +119,85 @@ function cmd_traceroute {
...
+
+# @sudo cmd_intrusion_show admin,tech
+# @restrict cmd_intrusion_show admin,tech
+# @doc cmd_intrusion_show
+# Show the chassis intrusion detection status
+function cmd_intrusion_show {
+  state=$(get_intrusion_state)
+  if [[ -z "$state" ]]; then
+    echo "Intrusion sensor service is stopped or failed"
+  else
```
nest1ing, line 165: Could you join it as `if ... elif ... else .. fi`?.

commands/diagnostics.vegman
```diff
@@ -119,3 +119,85 @@ function cmd_traceroute {
...
+  while [[ $# -gt 0 ]]; do
+    case "$1" in
+      -y | --yes) yes="y";;
+      *) abort_badarg "$1";;
+    esac
+    shift
+  done
+
+  if [[ -z "${yes}" ]]; then
+    confirm
```
nest1ing, line 191: I believe the code below should be moved outside of `if [[ -z "${yes}" ]]` statement or it won't be executed when `-y` specified.

nest1ing (dismissed):

commands/diagnostics.vegman
```diff
@@ -119,3 +119,84 @@ function cmd_traceroute {
...
+  fi
+
+  state=$(get_intrusion_state)
+  if [[ -z "$state" ]]; then
+    state="intrusion service is not enabled"
+  fi
+
+  pin=$(gpiofind ${INTR_RST_GPIO_PIN})
+  if [[ -n "${pin}" ]]; then
```
alatarum: I recommend to place this check to very beginning of the function (between while loop and "if -z ${yes}") and place some message like "intrusion sensor reset is not applicable for the platform"

AlexanderAmelkin: I agree. Let's make the message be "Intrusion detection reset is not available for this product"

commands/diagnostics.vegman
```diff
@@ -119,3 +119,84 @@ function cmd_traceroute {
...
+    shift
+  done
+
+  if [[ -z "${yes}" ]]; then
+    confirm
+  fi
+
+  state=$(get_intrusion_state)
+  if [[ -z "$state" ]]; then
+    state="intrusion service is not enabled"
```
AlexanderAmelkin: Let's make it "n/a" or "unknown" to make the audit message shorter

AlexanderAmelkin (changes requested):

AlexanderAmelkin (approved):

PR #60, opened 2022-01-19
host:virtualmedia: Move busctl to vm-cli

ikmsk10:
Current host virtualmedia implementation based on the `busctl` that
doesn't have a valid mechanism to transfer the `unix-fd` of UNIX pipes.
That is following to impossible to specify mounting credentials.
    
That patch brings a way to process virtual-media operations via the special `vm-cli` tool.
    
End-user-impact: Now the `host virtualmedia mount` command accepts credentials properly.

### DIFF
--- a/commands/host.vegman
+++ b/commands/host.vegman
@@ -167,38 +167,7 @@ function cmd_virtualmedia {
 # @doc cmd_virtualmedia_status
 # Get the status of virtual media
 function cmd_virtualmedia_status {
-  local -a usb_status_metadata=( "USB0;Proxy" "USB1;Legacy" "USB2;Direct" )
-
-  for usb_item_metadata in "${usb_status_metadata[@]}"
-  do
-    local id="${usb_item_metadata%%;*}"
-    local vm_type="${usb_item_metadata##*;}"
-    local usb_status=$(busctl get-property xyz.openbmc_project.VirtualMedia \
-                                        /xyz/openbmc_project/VirtualMedia/${vm_type}/${id} \
-                                        xyz.openbmc_project.VirtualMedia.Process Active)
-    local image=""
-    echo -ne "Virtual Media ($id) status: "
-    if [ "$usb_status" == "b true" ]; then
-      echo "mounted"
-      image=$(busctl get-property  xyz.openbmc_project.VirtualMedia \
-                                  /xyz/openbmc_project/VirtualMedia/${vm_type}/${id} \
-                                  xyz.openbmc_project.VirtualMedia.MountPoint ImageURL)
-      image=${image#*\"}
-      if [[ "${vm_type}" == "Direct" ]]; then
-        local export_name=""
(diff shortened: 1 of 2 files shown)


commands/host.vegman
```diff
@@ -280,36 +245,14 @@ function cmd_virtualmedia_mount {
...
-    busctl call xyz.openbmc_project.VirtualMedia \
-      /xyz/openbmc_project/VirtualMedia/Legacy/USB1 \
-      xyz.openbmc_project.VirtualMedia.Legacy \
-      Mount sbyv "$imagefile" $readwrite ${type} i ${cred_pipe} > /dev/null  && \
-    (echo "Virtual Media (USB1) mounted"; \
-    log_audit "Host virtual media mount : ${imagefile}") || \
-    echo "Failed to mount Virtual Media (USB1)"
+    vm-cli -o mount ${type} ${request_creds} ${readwrite} --${mode} --file ${imagefile} && \
+      log_audit "Host virtual media mount : ${imagefile}" || \
+      log_audit "Failed to mount Virtual Media (USB1)"
```
nest1ing: Are you sure that this and similars below message should be put in the audit log instead of user screen?
Look at the previous implementation:
- `log_audit` puts the message to the system logs with corresponding tags.
- `echo` puts the message to the user screen.

ikmsk10: The user message will be printed by the `vm-cli`.

nest1ing: And why have you decided to put the error messages to the audit logs?

ikmsk10: I believed the old implementation is missed that(error audit message) because in the umount operation we logging both case - success and failure.

commands/host.vegman
```diff
@@ -221,34 +190,30 @@ function cmd_virtualmedia_status {
...
   local imagefile_pattern="^(https?|smb|s?ftp|scp|nfs)://"
   local cred_pipe=0
   while [[ $# -gt 0 ]]; do
     case "${1}" in
-      --usb) type=0;;
-      --usb-ro) type=0;;
-      --hdd) type=1;;
-      --cdrom) type=2;;
-      -r) readwrite=false;;
+      --usb) type="--usb";;
```
alatarum: you can use something like
--usb|--usb-ro|--hdd|--cdrom) type="${1}";;

commands/host.vegman
nest1ing: No, the audit log should contain the facts of settings/mode changing.
The error messages are not relevant to it and will look like a trash.
I guess the previous umount implementation was just misreviewed.

ikmsk10: fixed

nest1ing (dismissed):
Looks good to me.
But I have some doubts about `vm-cli` which ones I will describe in the corresponding internal PR.
Thus, this PR likely will require some changes too.

alatarum (approved):

nest1ing (approved):

AlexanderAmelkin (approved):

PR #61, opened 2022-01-19
Add no-access role support

nest1ing:
The role `no-access` came from IPMI and requires a support.
The user with this role is unable to login.

### DIFF
--- a/commands/user
+++ b/commands/user
@@ -92,12 +92,13 @@ function cmd_show {
                                ${DBUS_ATTRIBUTES} "UserPrivilege" \
                                2>/dev/null \
           | awk '{gsub(/"/, "", $NF); print $NF}')
-    if [[ -z "${grp}" ]]; then
-      echo "Invalid user: ${user}" >&2
-      return 1
-    fi
+
     local role
-    role=$(group_to_role "${grp}")
+    if [[ -z "${grp}" ]]; then
+      role="no-access"
+    else
+      role=$(group_to_role "${grp}")
+    fi
     echo "${user}: ${role}"
   done
 }


ikmsk10 (commented):
Looks like `shellcheck` gives warnings that are we discussed at the TG chat and I believe current changes kindly should be based on the https://github.com/YADRO-KNS/obmc-yadro-cli/pull/60/commits/f736a0765c5f2348efdfcb7216f6fd2b3cf79990.
As far I can see the mentioned commit affects the same `commands/user` lines that are current PR, which will lead to conflict.

I'll just keep that note here to mark the required merge.

alatarum (approved):

commands/user
```diff
@@ -92,12 +92,13 @@ function cmd_show {
...
                                2>/dev/null \
           | awk '{gsub(/"/, "", $NF); print $NF}')
+
+    local role
     if [[ -z "${grp}" ]]; then
-      echo "Invalid user: ${user}" >&2
-      return 1
+      role="no-access"
+    else
+      role=$(group_to_role "${grp}")
```
AlexanderAmelkin, line 100: why don't you change the `group_to_role` subroutine instead?

nest1ing: Because I don't want to add the possibility to set this stupid role. This commit makes it printable only when it set by some other ways.

AlexanderAmelkin: How modifying `group_to_role` could aid in 'setting this stupid role' ? You're clearly amending the `group_to_role`'s behavior but doing it externally. I don't like it.

nest1ing:
This PR is no relevant more as we decided to remove `no-access` at all.

PR #63, opened 2022-05-19
clicmd: change the command name in the help info

Ramkosh:
Change the command name, displayed in the help
usage info section from 'bmc' to 'health'.

End-user-impact: Change the command name in the help usage info.

Signed-off-by: Kirill Pakhomov <k.pakhomov@yadro.com>

### DIFF
--- a/clicmd
+++ b/clicmd
@@ -78,7 +78,8 @@ fi
 # by the name of the CLI command that called this function
 function exec_tool {
   local FNAME="${FUNCNAME[1]##cmd_}"
-  local PNAME="bmc ${FNAME//_/ }"
+  local PNAME
+  PNAME="$(basename "${USER_CMD}") ${FNAME//_/ }"
   local TOOL="$1"; shift
   exec -a "${PNAME}" "${TOOL}" "$@"
 }


clicmd
```diff
@@ -78,7 +78,7 @@ fi
 # by the name of the CLI command that called this function
 function exec_tool {
   local FNAME="${FUNCNAME[1]##cmd_}"
-  local PNAME="bmc ${FNAME//_/ }"
+  local PNAME="health ${FNAME//_/ }"
```
AlexanderAmelkin: This looks wrong, just like the original code.
I looks like the original code assumed that `exec_tool` was only called from `bmc.*` scripts, which is not true now.
It looks like `$(basename "${USER_CMD}")` must be used instead of `bmc` or `health`.

AlexanderAmelkin (changes requested):

clicmd
```diff
@@ -78,7 +78,8 @@ fi
 # by the name of the CLI command that called this function
 function exec_tool {
   local FNAME="${FUNCNAME[1]##cmd_}"
-  local PNAME="bmc ${FNAME//_/ }"
+  local PNAME
```
Ramkosh, line 81: As recommended by [shellcheck](https://www.shellcheck.net/wiki/SC2155).

AlexanderAmelkin (approved):

PR #64, opened 2022-06-22
clicmd: fixed invalid regular expression when creating a new user

surf148:
Requirements:
  - Cannot start with a number
  - No special characters except underscore
These requirements are necessary due to the need for consistency with the WebUI

End-user-impact: fixed invalid regular expression when creating a new user
Signed-off-by: Evgeniy Kislenok <e.kislenok@yadro.com>

### DIFF
--- a/commands/user
+++ b/commands/user
@@ -19,7 +19,7 @@ DBUS_ROOT_PATH="/xyz/openbmc_project/user"
 DBUS_ATTRIBUTES="xyz.openbmc_project.User.Attributes"
 DBUS_DELETE="xyz.openbmc_project.Object.Delete"
 
-USERNAME_REGEX="^[\.a-zA-Z0-9_-]+$"
+USERNAME_REGEX="^[a-zA-Z_][a-zA-Z0-9_]*$"
 
 # This one us used in help, need to output a single word of role
 # and nothing else. Please do not 'beautify'.


commands/user
```diff
@@ -19,7 +19,7 @@ DBUS_ROOT_PATH="/xyz/openbmc_project/user"
 DBUS_ATTRIBUTES="xyz.openbmc_project.User.Attributes"
 DBUS_DELETE="xyz.openbmc_project.Object.Delete"
 
-USERNAME_REGEX="^[\.a-zA-Z0-9_-]+$"
+USERNAME_REGEX="^[a-zA-Z_][a-zA-Z0-9_]*$"
```
nest1ing, line 22: It's against RFC but fully according [WebUI](https://github.com/openbmc/webui-vue/blob/f763cd2e39ffce9b10191402243e8704794f08ff/src/views/SecurityAndAccess/UserManagement/ModalUser.vue#L296).

nest1ing (dismissed):

commands/user
AlexanderAmelkin: RFC is about generic POSIX systems. BMC may impose additional restrictions.

AlexanderAmelkin: It would be nice though to mention the source of those requirements/restrictions in the commit log.

AlexanderAmelkin (changes requested):

AlexanderAmelkin (approved):

PR #65, opened 2022-06-22
Implemented functional for enable/disable/unlock users in the CLI

surf148:
These functions supports multiple user names in one operation.
Function 'lock' is not implemented because lock status is calculated dynamically/

End-user-impact: Implement functional for enable/disable/unlock users in the CLI
Signed-off-by: Evgeniy Kislenok <e.kislenok@yadro.com>

### DIFF
--- a/commands/user
+++ b/commands/user
@@ -257,3 +257,80 @@ function cmd_set_role {
   log_audit "Set role ${role} for user ${name}"
   echo "User ${name} now has ${role} privileges."
 }
+
+# helper function for read and check user names
+function get_names {
+  local -a names
+
+  while [[ $# -gt 0 ]]; do
+    if [[ $1 =~ ${USERNAME_REGEX} ]]; then
+      names+=( "$1" )
+    else
+      abort_badarg "$1"
+    fi
+    shift
+  done
+
+  if [[ ${#names[@]} -lt 1 ]]; then
+    echo "User name is not specified" >&2
+    return 0
+  fi
+  echo "${names[@]}"
+}
+
+# @sudo cmd_unlock admin
+# @restrict cmd_unlock admin
+# @doc cmd_unlock
+# Unlock users
+#   USERNAME... - User account name (multiple names can be unlocked)
+function cmd_unlock {
+  local -a names
+  read -r -a names < <(get_names "$@")
+
+  if [[ ${#names[@]} -gt 0 ]]; then
+    for name in "${names[@]}"; do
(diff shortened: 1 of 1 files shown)


nest1ing (dismissed):

alatarum (dismissed):

nest1ing (dismissed):

commands/user
```diff
@@ -257,3 +257,80 @@ function cmd_set_role {
   log_audit "Set role ${role} for user ${name}"
   echo "User ${name} now has ${role} privileges."
 }
+
+# helper function for read and check user names
+function get_names {
+  local names=()
```
AlexanderAmelkin: It's better to not combine `local` with initialization. In this particular case it doesn't make any difference, but if you cared for the completion code of the initialization, then `local` on the same line would clobber it.

Also, if you're to define an array, please use `local -a` just like you would do with `declare -a`.

commands/user
```diff
@@ -257,3 +257,80 @@ function cmd_set_role {
...
+  echo "${names[@]}"
+}
+
+# @sudo cmd_unlock admin
+# @restrict cmd_unlock admin
+# @doc cmd_unlock
+# unlock users
+#   USERNAME... - User account name (multiple names can be unlocked)
+function cmd_unlock {
+  local names=()
```
AlexanderAmelkin: same here

commands/user
```diff
@@ -257,3 +257,80 @@ function cmd_set_role {
...
+      log_audit "Unlocked user ${name}"
+      echo "User ${name} unlocked."
+    done
+  fi
+}
+
+# @sudo cmd_enable admin
+# @restrict cmd_enable admin
+# @doc cmd_enable
+# enable users
```
AlexanderAmelkin: Please capitalize the first letter in command descriptions

commands/user
```diff
@@ -257,3 +257,80 @@ function cmd_set_role {
...
+      log_audit "Enabled user ${name}"
+      echo "User ${name} enabled."
+    done
+  fi
+}
+
+# @sudo cmd_disable admin
+# @restrict cmd_disable admin
+# @doc cmd_disable
+# disable users
```
AlexanderAmelkin: Here too

AlexanderAmelkin (changes requested):
Please obey the 50/72 rule in the commit log.
Also, please use the command name in the summary, e.g.:

```
vegman: user: Add disable/enable/unlock

Implemented functions to enable/disable/unlock users in the CLI
...
```

AlexanderAmelkin (approved):

AlexanderAmelkin (approved):

PR #66, opened 2022-06-24
Add syslog server address validation

Smith-x86:
Added:
1. Validity check of remote log server's IPv4-address
  (error message is displayed, busctl set-property isn't called);
2. Check for missing address[:port] parameter completely
  (if the parameter is missing, the port is set to 0)

Signed-off-by: Vladimir Kuznetsov <v.kuznetsov@yadro.com>


nest1ing (commented):
I suppose FQDN is also should be allowed here.
Likely it would be more correct to implement this management in `obmc-yadro-netconfig` with corresponding check and then rewrite this wrapper for using it.

alatarum:
> I suppose FQDN is also should be allowed here. Likely it would be more correct to implement this management in `obmc-yadro-netconfig` with corresponding check and then rewrite this wrapper for using it.

+1

alatarum (dismissed):

AlexanderAmelkin (approved):

PR #71, opened 2022-10-12
bmc datetime show: Remove the -u option

AlexanderAmelkin:
Remove the -u option as the time in BMC is always in UTC, and there is no way to set the time zone

End-user-impact: Fixed help for `bmc datetime show` command,
	             removed the useless `-u` option

### DIFF
--- a/commands/bmc.nicole
+++ b/commands/bmc.nicole
@@ -20,17 +20,12 @@ NETWORK_CLIENT_IFACE="xyz.openbmc_project.Network.Client"
 # Format date in a non-standard way
 function format_date {
   local compact="$1"
-  local utc="$2"
-  local sec="$3"
+  local sec="$2"
 
-  if [[ -n "${compact}" ]] && [[ -n "${utc}" ]]; then
-    date --date "@${sec}" --utc +'%F %T UTC'
-  elif [[ -n "${compact}" ]]; then
+  if [[ -n "${compact}" ]]; then
     date --date "@${sec}" +'%F %T UTC%z'
-  elif [[ -n "${utc}" ]]; then
-    date --date "@${sec}" --utc +'%a %d %b %Y %T UTC'
   else
-    date --date "@${sec}" +'%a %d %b %Y %T UTC%z'
+    date --date "@${sec}" --utc +'%a %d %b %Y %T UTC'
   fi
 }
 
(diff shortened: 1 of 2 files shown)


commands/bmc.nicole
```diff
@@ -20,17 +20,12 @@ NETWORK_CLIENT_IFACE="xyz.openbmc_project.Network.Client"
 # Format date in a non-standard way
 function format_date {
   local compact="$1"
-  local utc="$2"
   local sec="$3"
 
-  if [[ -n "${compact}" ]] && [[ -n "${utc}" ]]; then
```
ikmsk10, line 26: I can't understand this change, Is it will be 
```bash
local sec="$3"

elif [[ -n "${compact}" ]]; then
    date --date "@${sec}" +'%F %T UTC%z'
...
```
It seems to me that this is incorrect. This should be at least

```bash
if [[ -n "${compact}" ]]; then
    date --date "@${sec}" +'%F %T UTC%z'
    
```

AlexanderAmelkin: Thanks. Typo indeed. Changed in one place and forgot here. It must be `if`.

AlexanderAmelkin: Fixed it

commands/bmc.nicole
```diff
@@ -20,17 +20,12 @@ NETWORK_CLIENT_IFACE="xyz.openbmc_project.Network.Client"
 # Format date in a non-standard way
 function format_date {
   local compact="$1"
-  local utc="$2"
   local sec="$3"
```
nest1ing: Likely `sec` must be initialized with `"$2"` now.

commands/bmc.vegman
```diff
@@ -16,17 +16,12 @@ TIMESYNC_METHOD_PROPERTY="TimeSyncMethod"
 # Format date in a non-standard way
 function format_date {
   local compact="$1"
-  local utc="$2"
-  local sec="$3"
+  local sec="$2"
```
nest1ing, line 19: Oh, You have done it here, but forgot there. :)

nest1ing (changes requested):

commands/bmc.nicole
AlexanderAmelkin: Fixed

nest1ing (approved):

PR #72, opened 2022-10-12
vegman: diagnostics: Add intrusion reset request

Ramkosh:
Added functionality to support intrusion sensor status reset request on VEGMAN-R GEN1 hardware.

End-user-impact: Added functionality to support
                 intrusion sensor status reset request
                 on VEGMAN-R GEN1 hardware.

Signed-off-by: Kirill Pakhomov <k.pakhomov@yadro.com>

### DIFF
--- a/commands/diagnostics.vegman
+++ b/commands/diagnostics.vegman
@@ -129,6 +129,12 @@ INTR_DBUS_SERVICE="xyz.openbmc_project.IntrusionSensor"
 INTR_DBUS_PATH="/xyz/openbmc_project/Intrusion/Chassis_Intrusion"
 INTR_DBUS_INTERFACE="xyz.openbmc_project.Chassis.Intrusion"
 INTR_DBUS_PROPERTY="Status"
+# D-Bus Intrusion sensor reset request status objects
+INTR_REQ_STATUS_DBUS_PATH="/xyz/openbmc_project/Intrusion/Reset_Request"
+INTR_REQ_STATUS_DBUS_INTF="xyz.openbmc_project.Chassis.Intrusion.ResetRequest"
+INT_REQ_STATUS_DBUS_PROPERTY="ClearRequestStatus"
+PROPERTY_PENDING="Pending"
+GEN_SMI_GPIO_PIN="IRQ_BMC_PCH_SMI_LPC"
 
 # @doc cmd_intrusion
 # Chassis intrusion detection commands
@@ -180,6 +186,27 @@ get_intrusion_service() {
          echo Enabled || echo Disabled
 }
 
+get_intrusion_request_status() {
+  local state
+  state=$(busctl get-property ${INTR_DBUS_SERVICE} "${INTR_REQ_STATUS_DBUS_PATH}" \
+                      ${INTR_REQ_STATUS_DBUS_INTF} ${INT_REQ_STATUS_DBUS_PROPERTY} \
+                      2>/dev/null | awk '{print $NF;}' ||:)
+  if [[ "${state}" == "\"${PROPERTY_PENDING}\"" ]]; then
+    echo "(clear pending)"
+  fi
+}
+
+set_intrusion_request_status() {
+  local value=$1
+  busctl set-property ${INTR_DBUS_SERVICE} "${INTR_REQ_STATUS_DBUS_PATH}" \
+                      ${INTR_REQ_STATUS_DBUS_INTF} ${INT_REQ_STATUS_DBUS_PROPERTY} \
+                      "s" "${value}" \
+                      2>/dev/null | awk '{print $NF;}' ||:
+
+  echo "Set intrusion sensor request status to: ${value}"
+  log_audit "Set intrusion sensor request status to: ${value}"
+}
+
 # @sudo cmd_intrusion_status admin,operator,user
 # @doc cmd_intrusion_status
 # Show the chassis intrusion detection status
@@ -191,7 +218,8 @@ function cmd_intrusion_status {
 	  return 0
   fi
 
-  echo "Chassis intrusion state: $(get_intrusion_state)"
+  sstate="$(get_intrusion_request_status)"
+  echo "Chassis intrusion state: $(get_intrusion_state) ${sstate}"
 }
 
 # @sudo cmd_intrusion_clear admin
@@ -219,8 +247,28 @@ function cmd_intrusion_clear {
   pin=$2
 
   if [[ -z "${chip}" ]]; then
-    echo "Intrusion detection reset is not available for this product"
-    return 1
+    # Check for IRQ_BMC_PCH_SMI_LPC pin on VEGMAN-R hardware
+    set -- $(gpiofind ${GEN_SMI_GPIO_PIN}) "" ""
+    chip=$1
+    pin=$2
+    if [[ -z "${chip}" ]]; then
+      echo "Intrusion detection reset is not available for this product"
+      return 1
+    fi
+
+    cmd_intrusion_status
+
+    if [[ -z "${yes}" ]]; then
+      echo -n "Request intrusion sensor state reset? "
+      confirm
+    fi
+
+    set_intrusion_request_status ${PROPERTY_PENDING}
+    # generate SMI
+    gpioset "${chip}" "${pin}"=0
+    gpioset "${chip}" "${pin}"=1
+    cmd_intrusion_status
+    return 0
   fi
 
   local state


commands/diagnostics.vegman
```diff
@@ -180,6 +187,31 @@ get_intrusion_service() {
...
+    echo "N/A"
+  elif [[ "${state}" == "\"Cleared\"" ]]; then
+    echo "Reset intrusion sensor state request: Cleared"
+  else
+    echo "Reset intrusion sensor state request: Pending"
+  fi
+}
+
+set_intrusion_request_status() {
+  local property=$1
```
AlexanderAmelkin: Isn't it `value` ?

commands/diagnostics.vegman
```diff
@@ -218,12 +250,31 @@ function cmd_intrusion_clear {
...
+    pin=$2
+    if [[ -z "${chip}" ]]; then
+      echo "Intrusion detection reset is not available for this product"
+      return 1
+    fi
+    state=$(get_intrusion_state)
+    echo "Current chassis intrusion state: $(get_intrusion_state)"
+    state=$(get_intrusion_request_status)
+    echo "Current ${state}"
+    echo -n "Request intrusion sensor state reset? "
```
AlexanderAmelkin: This must only be asked if no `-y|--yes` option is provided. Check how that is done below for other vegmans.

commands/diagnostics.vegman
```diff
@@ -218,12 +250,31 @@ function cmd_intrusion_clear {
...
+    chip=$1
+    pin=$2
+    if [[ -z "${chip}" ]]; then
+      echo "Intrusion detection reset is not available for this product"
+      return 1
+    fi
+    state=$(get_intrusion_state)
+    echo "Current chassis intrusion state: $(get_intrusion_state)"
+    state=$(get_intrusion_request_status)
+    echo "Current ${state}"
```
AlexanderAmelkin: I guess `${state,}` would work better here yielding `Current reset...` instead of `Current Reset...` as it is now.

commands/diagnostics.vegman
```diff
@@ -180,6 +187,31 @@ get_intrusion_service() {
...
+  state=$(busctl set-property ${INTR_DBUS_SERVICE} "${INTR_REQ_STATUS_DBUS_PATH}" \
+                      ${INTR_REQ_STATUS_DBUS_INTF} ${INT_REQ_STATUS_DBUS_PROPERTY} \
+                      "s" "${property}" \
+                      2>/dev/null | awk '{print $NF;}' ||:)
+
+  echo "Set intrusion sensor request status to:" "${property}"
+}
+
 # @sudo cmd_intrusion_status admin,operator,user
 # @doc cmd_intrusion_status
```
AlexanderAmelkin, line 211: I believe this command also needs to report the pending request state

AlexanderAmelkin (changes requested):

commands/diagnostics.vegman
Ramkosh: then I'll remove the `echo "Current ${state}"` from the `function cmd_intrusion_clear`, otherwise we'll just have the same line displayed twice.

Ramkosh:
added a couple of empty lines to space out logic blocks in order to improve readability

Ramkosh:
- [x] forgot to say: i'm yet update the dbus interface in this PR

commands/diagnostics.vegman
AlexanderAmelkin: Ok, but I still don't see it here

commands/diagnostics.vegman
```diff
@@ -218,12 +251,34 @@ function cmd_intrusion_clear {
...
+
+    if [[ -z "${yes}" ]]; then
+      echo -n "Request intrusion sensor state reset? "
+      confirm
+    fi
+
+    set_intrusion_request_status ${PROPERTY_PENDING}
+    # generate SMI
+    gpioset "${chip}" "${pin}"=0
+    gpioset "${chip}" "${pin}"=1
```
nest1ing, line 269: Use `gpioset --mode time --sec/usec ...` instead of two calls.

nest1ing: Oh, no. In this case the pin keeps the last state.
Let it be as is.

commands/diagnostics.vegman
```diff
@@ -180,6 +187,31 @@ get_intrusion_service() {
...
+  elif [[ "${state}" == "\"Pending\"" ]]; then
+    echo "clear request pending"
+  else
+    echo "no clear request pending"
+  fi
+}
+
+set_intrusion_request_status() {
+  local value=$1
+  local state
```
nest1ing: state looks like unused. Is it really required?

commands/diagnostics.vegman
```diff
@@ -180,6 +187,31 @@ get_intrusion_service() {
...
 }
 
+get_intrusion_request_status() {
+  local state
+  state=$(busctl get-property ${INTR_DBUS_SERVICE} "${INTR_REQ_STATUS_DBUS_PATH}" \
+                      ${INTR_REQ_STATUS_DBUS_INTF} ${INT_REQ_STATUS_DBUS_PROPERTY} \
+                      2>/dev/null | awk '{print $NF;}' ||:)
+  if [[ -z "${state}" ]]; then
+    echo "clear request state N/A"
+  elif [[ "${state}" == "\"Pending\"" ]]; then
```
nest1ing: it would be better to use `${PROPERTY_PENDING}` instead of hardcode.

commands/diagnostics.vegman
```diff
@@ -129,6 +129,13 @@ INTR_DBUS_SERVICE="xyz.openbmc_project.IntrusionSensor"
 INTR_DBUS_PATH="/xyz/openbmc_project/Intrusion/Chassis_Intrusion"
 INTR_DBUS_INTERFACE="xyz.openbmc_project.Chassis.Intrusion"
 INTR_DBUS_PROPERTY="Status"
+# D-Bus Intrusion sensor reset request status objects
+INTR_REQ_STATUS_DBUS_PATH="/xyz/openbmc_project/Intrusion/Reset_Request"
+INTR_REQ_STATUS_DBUS_INTF="xyz.openbmc_project.Chassis.Intrusion.ResetRequest"
+INT_REQ_STATUS_DBUS_PROPERTY="ClearRequestStatus"
+PROPERTY_CLEARED="No request"
```
nest1ing: `shellcheck` warns at this line:
```
 SC2034: PROPERTY_CLEARED appears unused.
 ```

File: commands/diagnostics.vegman [hunk 9]
```diff
@@ -180,6 +186,30 @@ get_intrusion_service() {
...
+}
+
+set_intrusion_request_status() {
+  local value=$1
+  busctl set-property ${INTR_DBUS_SERVICE} "${INTR_REQ_STATUS_DBUS_PATH}" \
+                      ${INTR_REQ_STATUS_DBUS_INTF} ${INT_REQ_STATUS_DBUS_PROPERTY} \
+                      "s" "${value}" \
+                      2>/dev/null | awk '{print $NF;}' ||:
+
+  echo "Set intrusion sensor request status to:" "${value}"
```
nest1ing: Likely we should post the same message to the audit log.
I suggest to add here something like:
```sh
log_audit "Set intrusion sensor request status to: ${value}"
```

BTW: Why have You split a message string and `${value}`? ;)

nest1ing (dismissed):

commands/diagnostics.vegman
Ramkosh: 1) should we log here or intel-ipmi-oem?
2) must've missed bracket auto-completion or something, thanks!

nest1ing (dismissed):

commands/diagnostics.vegman
nest1ing: I believe it should be in both places.
1-th will log the commands sent via IPMI
2-th via CLI.

Ramkosh: Right. Added the audit logging here and enabled it in the IPMI audit header (that should log IPMI cmds automatically).

commands/diagnostics.vegman
```diff
@@ -191,7 +222,8 @@ function cmd_intrusion_status {
 	  return 0
   fi
 
-  echo "Chassis intrusion state: $(get_intrusion_state)"
+  sstate="$(get_intrusion_request_status)"
```
AlexanderAmelkin, line 221: So, now instead of
`Chassis intrusion state: Intrusion detected`
the user will be getting
`no clear request pending` ?

That's definitely not what is needed or what we discussed.

Please make sure that, as we discussed, for `diagnostics intrusion status` the users gets these, depending on the circumstances:

`Chassis intrusion state: Intrusion detected`
`Chassis intrusion state: Intrusion detected (clear pending)`
`Chassis intrusion state: No intrusion`
`Chassis intrusion state: N/A`

commands/diagnostics.vegman
```diff
@@ -218,12 +250,34 @@ function cmd_intrusion_clear {
...
+    set -- $(gpiofind ${GEN_SMI_GPIO_PIN}) "" ""
+    chip=$1
+    pin=$2
+    if [[ -z "${chip}" ]]; then
+      echo "Intrusion detection reset is not available for this product"
+      return 1
+    fi
+
+    state=$(get_intrusion_state)
+    echo "Сhassis intrusion state: $(get_intrusion_state)"
```
AlexanderAmelkin: The same as in `cmd_intrusion_status`

commands/diagnostics.vegman
```diff
@@ -218,12 +250,34 @@ function cmd_intrusion_clear {
...
+      echo -n "Request intrusion sensor state reset? "
+      confirm
+    fi
+
+    set_intrusion_request_status ${PROPERTY_PENDING}
+    # generate SMI
+    gpioset "${chip}" "${pin}"=0
+    gpioset "${chip}" "${pin}"=1
+    state=$(get_intrusion_request_status)
+    echo "Сhassis intrusion state: $(get_intrusion_state) (${sstate})"
```
AlexanderAmelkin: The same as in `cmd_intrusion_status`

AlexanderAmelkin (changes requested):

commands/diagnostics.vegman
Ramkosh: We will be getting the following:
`Chassis intrusion state: %Intrusion status% (%reset_request_status%)`
where %reset_request_status% can be one of the following:
`clear request state N/A`
`clear request pending`
`no clear request pending`

UPD: re-doing this as per agreement.

AlexanderAmelkin (approved):

